PR #11, opened 2017-07-13
Larmor X-Y Beamstop

samueljackson92:
This adds a suite of tests for checking the correctness of the Larmor X-Y beamstop

The issue for this is: ISISComputingGroup/IBEX/issues/2437
The main PR is: ISISComputingGroup/EPICS-ioc/pull/160


tests/xybeamstop.py
```diff
@@ -0,0 +1,164 @@
...
+    def _set_pv_value(self, pv_name, value):
+        self.ca.set_pv_value("{0}:{1}".format(PREFIX, pv_name), value)
+
+    def _set_x(self, value):
+        self._set_pv_value("ARM:X", value)
+
+    def _set_y(self, value):
+        self._set_pv_value("ARM:Y", value)
+
+    def test_GIVEN_x_y_WHEN_read_x_y_THEN_x_y_is_as_expected(self):
```
AdrianPotter: This test name is too generic. You should never use `as_expected` in a test name. The test name should tell you what is expected. It's akin to saying `test_result_correct`.  There seems to be some confusion about the role of GIVEN-WHEN-THEN as well.

GIVEN: The test context. What state is the device in? In this case, there isn't any special state so it can be omitted
WHEN: The action that is taken. The action here isn't reading x and y, it's setting x and y
THEN: The expected result. Here the expected result is that the device has reached its expected position

So something along the lines of:

```
test_WHEN_the_x_y_position_is_set_THEN_the_beam_stop_moves_to_the_supplied_position
```

**Please fix all of the other test names too**

tests/xybeamstop.py
```diff
@@ -0,0 +1,164 @@
...
+        y = 2.0
+        self._set_x(x)
+        time.sleep(2)
+        self._set_y(y)
+        time.sleep(4)
+
+        self.ca.assert_that_pv_is_close("MOT:ARM:X", x, 1e-2)
+        self.ca.assert_that_pv_is_close("MOT:ARM:Y", y, 1e-2)
+        self.ca.assert_that_pv_is_close("MOT:ARM:X.RBV", x, 1e-1)
+        self.ca.assert_that_pv_is_close("MOT:ARM:Y.RBV", y, 1e-1)
```
AdrianPotter: The above test introduces un-commented, arbitrary, different wait times for moving to the different positions. I suggest using (or implementing if it doesn't exist) a timeout on the asserts instead.

**Please remove all subsequent arbitrary waits too**

AdrianPotter: Should this be the `RBV` or `SP:RBV`? Please could you explain why `MOT:ARM:X` and `MOT:ARM:X.RBV` are different, why we need both, and why they need different tolerances.

tests/xybeamstop.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def test_GIVEN_store_command_WHEN_read_x_y_is_as_expected(self):
```
AdrianPotter: GIVEN can sometimes be omitted but all tests should have WHEN and THEN. Please see the explanation above and update the test name appropriately

tests/xybeamstop.py
```diff
@@ -0,0 +1,164 @@
...
+        self._set_pv_value("ARM:STORE", 1)
+        time.sleep(1)
```
AdrianPotter: Use a timeout on the subsequent assert rather than introducing an arbitrary timeout

tests/xybeamstop.py
```diff
@@ -0,0 +1,164 @@
...
+
+        self.ca.assert_that_pv_is("MOT:ARM:STORE", "STORED")
+        self.ca.assert_that_pv_is_close("MOT:ARM:X", -7.071)
+        self.ca.assert_that_pv_is_close("MOT:ARM:Y", 2.929)
+        self.ca.assert_that_pv_is_close("MOT:ARM:X.RBV", -7.071, 1e-2)
+        self.ca.assert_that_pv_is_close("MOT:ARM:Y.RBV", 2.929, 1e-2)
```
AdrianPotter: As above, I don't understand why RBV and non-RBV values need to be checked independently and why they need different tolerances

tests/xybeamstop.py
```diff
@@ -0,0 +1,164 @@
...
+        self.ca.assert_that_pv_is_close("MOT:ARM:X.RBV", x, 1e-1)
+        self.ca.assert_that_pv_is_close("MOT:ARM:Y.RBV", y, 1e-1)
+
+    def test_GIVEN_store_command_WHEN_read_x_y_is_as_expected(self):
+        self._set_pv_value("ARM:STORE", 1)
+        time.sleep(1)
+
+        self.ca.assert_that_pv_is("MOT:ARM:STORE", "STORED")
+        self.ca.assert_that_pv_is_close("MOT:ARM:X", -7.071)
+        self.ca.assert_that_pv_is_close("MOT:ARM:Y", 2.929)
```
AdrianPotter: These positions `-7.071` and `2.929` are used throughout:

1. Where have they come from? It's not clear from the test file why this should be the expected position (a good example of why `as_expected` should never be used in test naming.
1. As the same value is used repeatedly throughout the set of tests, it should be refactored into a class variable, rather than appearing as a magic number

**Please remove all subsequent magic numbers too**

tests/xybeamstop.py
AdrianPotter: The same PVs are referenced throughout the test set. Use a variable instead to avoid duplication.

tests/xybeamstop.py
```diff
@@ -0,0 +1,164 @@
...
+    #     self.ca.assert_that_pv_is("MOT:ARM:Y", y_lower)
+    #     self.ca.assert_that_pv_is_close("MOT:ARM:Y.RBV", y_lower, 1e-1)
+    #
+    # def test_GIVEN_set_y_outside_upper_limit_WHEN_read_y_is_not_outside_limit(self):
+    #     y_upper = 2.929
+    #     self._set_y(y_upper)
+    #     self._set_y(y_upper + 1.0)
+    #     time.sleep(1)
+    #     self.ca.assert_that_pv_is_close("MOT:ARM:Y", y_upper)
+    #     self.ca.assert_that_pv_is_close("MOT:ARM:Y.RBV", y_upper, 1e-2)
```
AdrianPotter: Are these tests not being run for a reason? Please either reactivate them or get rid of them

utils/channel_access.py
```diff
@@ -124,6 +124,26 @@ def assert_that_pv_is_an_integer_between(self, pv, min, max, timeout=None):
...
 
+    def assert_that_pv_is_close(self, pv, expected_value, delta=1e-6, timeout=None):
+        """
+        Assert that the pv has one of the expected values or that it becomes one of the expected value within the
+        timeout
+
+        :param pv: pv name
+        :param expected_value: value to compare to
+        :param delta: maximum acceptable absolute difference between the two values
+        :param timeout: if it hasn't changed within this time raise assertion error
```
AdrianPotter: Please state the units of time

utils/channel_access.py
```diff
@@ -124,6 +124,26 @@ def assert_that_pv_is_an_integer_between(self, pv, min, max, timeout=None):
...
+        :return:
+        :raises AssertionError: if value does not become within tolerance of the requested value
+        :raises UnableToConnectToPVException: if pv does not exist within timeout
+        """
+        error_message = self._wait_for_pv_lambda(lambda: self._value_is_close(pv, expected_value, delta), timeout)
+
+        if error_message is None:
+            return
+
+        raise AssertionError(error_message)
```
AdrianPotter: If possible, `return` should be the last thing a function does. Sometimes we sneakily do it early but it's not especially good practice. This would be better as:

```
if error_message is not None:
    raise ...
```

You then don't even need the `return` statement as it's implcit

utils/channel_access.py
```diff
@@ -140,6 +160,21 @@ def _values_match(self, pv, expected_value):
...
+        :param pv: name of the pv (no prefix)
+        :param expected_value: value that is expected
+        :param delta: maximum acceptable absolute difference between the values
+        :return: None if they match; error string stating the difference if they do not
+        """
+        pv_value = self.get_pv_value(pv)
+        if abs(pv_value - expected_value) < delta:
+            return None
+        else:
+            return "Expected {expected}: actual {actual}".format(expected=expected_value, actual=pv_value)
```
AdrianPotter: As above. You don't need the `if` here:

```
if not ... :
    return "Expected...
```

If the condition of the `if` isn't met, the function will get to the end and `return None` anyway.

run_all_tests.bat
```diff
@@ -50,8 +50,13 @@ echo;
 echo ---------------------------------------
 echo TESTING INSTRON Rec Sim
 REM call %PYTHON% "%EPICS_KIT_ROOT%\support\IocTestFramework\master\run_tests.py" -r -pf %MYPVPREFIX%  -d instron_stress_rig -p %EPICS_KIT_ROOT%\ioc\master\INSTRON\iocBoot\iocINSTRON-IOC-01
+echo ---------------------------------------
+echo;
 
 echo ---------------------------------------
+echo TESTING XYBEAMSTOP Dev Sim
+call %PYTHON% "%EPICS_KIT_ROOT%\support\IocTestFramework\master\run_tests.py" -pf %MYPVPREFIX%  -d xybeamstop -p %EPICS_KIT_ROOT%\ioc\master\GALIL\iocBoot\iocGALIL-IOC-01
```
AdrianPotter: This command doesn't actually result in the tests running. I suspect because it doesn't launch a GALIL IOC so the test has nothing to act on.

PR #19, opened 2017-08-09
Ticket2469 ieg

Tom-Willemsen:
IOC test framework tests for the IEG system

### DIFF
--- a/tests/ieg.py
+++ b/tests/ieg.py
@@ -8,6 +8,11 @@ from utils.testing import get_running_lewis_and_ioc
 from time import sleep
 
 
+CALIBRATION_A = 1.23
+CALIBRATION_B = -4.56
+
+MACROS = { "CALIBRATION_A": str(CALIBRATION_A), "CALIBRATION_B": str(CALIBRATION_B)}
+
 class IegTests(unittest.TestCase):
     """
     Tests for the IEG IOC.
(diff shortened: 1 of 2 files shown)


tests/ieg.py
```diff
@@ -0,0 +1,153 @@
...
+    def test_WHEN_valve_states_are_changed_on_device_THEN_valve_state_pv_updates(self):
+        for gas_valve_state in [True, False]:
+            self._lewis.backdoor_set_on_device("gas_valve_open", gas_valve_state)
+            for buffer_valve_state in [True, False]:
+                self._lewis.backdoor_set_on_device("buffer_valve_open", buffer_valve_state)
+                for pump_valve_state in [True, False]:
+                    self._lewis.backdoor_set_on_device("pump_valve_open", pump_valve_state)
+                    self.ca.assert_that_pv_is_number("IEG_01:VALVESTATE.B0", 1 if pump_valve_state is True else 0)
+                    self.ca.assert_that_pv_is_number("IEG_01:VALVESTATE.B1", 1 if buffer_valve_state is True else 0)
+                    self.ca.assert_that_pv_is_number("IEG_01:VALVESTATE.B2", 1 if gas_valve_state is True else 0)
```
AdrianPotter: Is `is True` necessary here? Given these values are explicitly boolean, perhaps `..._state` is the wrong choice of variable name. As we're only using `True` and `False`, you can just use `if *_state`.

Tom-Willemsen:
Can't reproduce test failure

PR #22, opened 2017-08-23
Ticket651 hifi cryomag psu test files

esouthren:


esouthren:
Pull Requests related to this Ticket:
[IOC Files](https://github.com/ISISComputingGroup/EPICS-ioc/pull/172)
[IOC Support Files](https://github.com/ISISComputingGroup/EPICS-Hifi_Cryomagnet_PSU/pull/1)
[Emulator](https://github.com/ISISComputingGroup/EPICS-DeviceEmulator/pull/34)
[Test Files](https://github.com/ISISComputingGroup/EPICS-IOC_Test_Framework/pull/22)
[OPI](https://github.com/ISISComputingGroup/ibex_gui/pull/542)

Note: globals.txt may need the following macros to run the tests:

```
# comment out first two lines for unit testing

#HFMAGPSU_01__EMULATOR_PORT=57677
#HFMAGPSU_01__DEVSIM=1
HFMAGPSU_01__PORT=COM20
```

tests/hfmagpsu.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def setUp(self):
+        self._lewis, self._ioc = get_running_lewis_and_ioc("HFMAGPSU") # emulator name
+        self.ca = ChannelAccess(default_timeout=10)
+        self.ca.wait_for("HFMAGPSU_01:DIRECTION", timeout=10)
+
+    def _write_direction(self, direction):
+        self._lewis.backdoor_set_on_device("direction", direction)
+        self.ca.set_pv_value("HFMAGPSU_01:LOGMESSAGE", "HH:MM:SS DIRECTION: [" + str(direction) + "]")
+        self.ca.set_pv_value("HFMAGPSU_01:DIRECTION:SP", direction)
```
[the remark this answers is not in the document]
samueljackson92: Actually, this is a mix of test "setup" code and "acting" code. The lewis backdoor sets the value on the device directly, bypassing the emulator. Therefore we can check the read works without doing the `self.ca.set_pv_value` call. We should then have another set of tests for check we can write to a set point.

tests/hfmagpsu.py
```diff
@@ -0,0 +1,122 @@
...
+        expected_output = 35.5
+        self._write_limit(expected_output)
+        self.ca.assert_that_pv_is("HFMAGPSU_01:LIMIT", expected_output)
+
+    def test_GIVEN_heaterValue_set_WHEN_read_THEN_heaterValue_is_as_expected(self):
+        expected_heaterValue = 15.5
+        self._write_heater_value(expected_heaterValue)
+        self.ca.assert_that_pv_is("HFMAGPSU_01:HEATERVALUE", expected_heaterValue)
+        
+    def test_GIVEN_pause_set_WHEN_read_THEN_pause_is_as_expected(self):
```
samueljackson92: This occasionally fails for me. After a bit of digging I found that the cause is because we expect to set the pause value via the backdoor to 'ON' but we're actually passing it `1`. This occasionally works because we are setting ti both through the backdoor and through the IOC set point, which causes a race condition which occasionally allows the test to pass.

tests/hfmagpsu.py
```diff
@@ -0,0 +1,122 @@
...
+        self._write_output_mode(0)
+        expected_message = "HH:MM:SS UNITS: [0]"
+        self.ca.assert_that_pv_is("HFMAGPSU_01:LOGMESSAGE", expected_message)  
+
+    def test_GIVEN_limit_set_WHEN_read_then_logmessage_is_as_expected(self):
+        self._write_limit(33.4)
+        expected_message = "HH:MM:SS VOLTAGE LIMIT: [33.4]"
+        self.ca.assert_that_pv_is("HFMAGPSU_01:LOGMESSAGE", expected_message)
+
+    def test_GIVEN_rampTarget_set_WHEN_read_THEN_rampTarget_is_as_expected(self):
```
samueljackson92: I found that this test would occasionally fail for the same reason as the `pause` test above. Please correct this the same way & please check and fix any other tests that might have this issue.

tests/hfmagpsu.py
esouthren: @Tom-Willemsen found an underlying bug here - when commands begin the same (e.g. `P` for readPause in this case) but one has additional arguments (eg `P ON` for writePause), sometimes Cmdbuilder will find the P, think it's got a command match, and build that without the arguments. He helped write a fix in command_builder.py that forces CmdBuilder to match the entire command.

esouthren: I've run the tests from other emulators that use CmdBuilder to make sure it hasn't stuffed them up. So far so good

tests/hfmagpsu.py
```diff
@@ -0,0 +1,102 @@
...
+class HfmagpsuTests(unittest.TestCase):
+
+    def setUp(self):
+        self._lewis, self._ioc = get_running_lewis_and_ioc("HFMAGPSU", DEVICE_PREFIX)
+        self.ca = ChannelAccess(device_prefix=DEVICE_PREFIX, default_timeout=30)
+        self.ca.wait_for("DIRECTION")
+        time.sleep(8)
+
+    def test_GIVEN_midTarget_set_WHEN_read_THEN_midTarget_is_as_expected(self):
+        test_value = 2.325
```
Rory-Potter-Tessella: Are your this test value representative of values the PV will take? Does it work with exponential notation e.g. `2e-2`? 

You can use [parameterized](https://github.com/wolever/parameterized) to parameterize your tests. Just pip install parameterized. This is going to be added to genie python install script (see https://github.com/ISISComputingGroup/genie_python/pull/135).

tests/hfmagpsu.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def test_GIVEN_constant_set_WHEN_read_THEN_constant_is_as_expected(self):
+        test_value = 0.0032
+        self.ca.assert_setting_setpoint_sets_readback(test_value, "CONSTANT")
+
+    def test_GIVEN_heaterValue_set_WHEN_read_THEN_heaterValue_is_as_expected(self):
+        test_value = 15.5
+        self.ca.assert_setting_setpoint_sets_readback(test_value, "HEATERVALUE")
+
+    def test_GIVEN_pause_set_ON_WHEN_read_THEN_pause_is_as_expected(self):
```
Rory-Potter-Tessella: After changing Values to an Enum this would become 

```
expected_output = Values.ON.name
set_value = Values.ON.value
```

utils/ioc_launcher.py
```diff
@@ -147,6 +147,7 @@ def open(self):
         self._process = subprocess.Popen(ioc_run_commandline, creationflags=subprocess.CREATE_NEW_CONSOLE,
                                          cwd=self._directory, stdin=subprocess.PIPE, stdout=self._logFile,
                                          stderr=subprocess.STDOUT, env=settings)
+
```
Rory-Potter-Tessella: Why was this line added? It is probably best to leave ioc_launcher the same.

tests/hfmagpsu.py
```diff
@@ -0,0 +1,102 @@
...
+        expected_output = StringValues.ON
+        set_value = EnumValues.ON
+        self.ca.assert_setting_setpoint_sets_readback(set_value, "HEATERSTATUS", "HEATERSTATUS:SP", expected_output)
+
+    def test_GIVEN_heaterStatus_set_OFF_WHEN_read_THEN_heaterStatus_is_as_expected(self):
+        expected_output = StringValues.OFF
+        set_value = EnumValues.OFF
+        self.ca.assert_setting_setpoint_sets_readback(set_value, "HEATERSTATUS", "HEATERSTATUS:SP", expected_output)
+
+    def test_GIVEN_rampTarget_set_WHEN_read_THEN_rampTarget_is_as_expected(self):
```
Rory-Potter-Tessella: You can use [parameterized ](https://github.com/wolever/parameterized) to run over a set of cases. Just pip install parameterized. This is going to be added to genie python install script  (see https://github.com/ISISComputingGroup/genie_python/pull/135).

tests/hfmagpsu.py
```diff
@@ -0,0 +1,102 @@
...
+TEST_MODES = [TestModes.DEVSIM, TestModes.RECSIM]
+
+
+class HfmagpsuTests(unittest.TestCase):
+
+    def setUp(self):
+        self._lewis, self._ioc = get_running_lewis_and_ioc("HFMAGPSU", DEVICE_PREFIX)
+        self.ca = ChannelAccess(device_prefix=DEVICE_PREFIX, default_timeout=30)
+        self.ca.wait_for("DIRECTION")
+        time.sleep(8)
```
Rory-Potter-Tessella: Do you need to sleep? This feels unnecessary.

PR #27, opened 2017-09-13
Ticket647 hifi cryo mag keithley 2700 scanner ioc

Mukz95:
IOC Tests for keithley2700 https://waffle.io/ISISComputingGroup/IBEX/cards/56cf2ecbca5cd1160040097f

### DIFF
--- a/tests/keithley_2700.py
+++ b/tests/keithley_2700.py
@@ -1,0 +1,188 @@
+import unittest
+from unittest import skipIf
+from utils.channel_access import ChannelAccess
+from utils.testing import get_running_lewis_and_ioc
+from utils.ioc_launcher import IOCRegister
+
+
+
+class Keithley_2700Tests(unittest.TestCase):
+    """
+    Tests for the Keithley2700.
+    """
+
+    def setUp(self):
+        self._lewis, self._ioc = get_running_lewis_and_ioc("keithley_2700")
+        self.ca = ChannelAccess(default_timeout=30)
+        self.ca.wait_for("KHLY2700_01:IDN")
+
+    @skipIf(IOCRegister.uses_rec_sim, "In rec sim this test fails")
+    def test_GIVEN_idn_defined_WHEN_read_THEN_idn_is_as_expected(self):
+        self.ca.assert_that_pv_is("KHLY2700_01:IDN", "Keithley 2700 Multimeter emulator.")
+
+    @skipIf(IOCRegister.uses_rec_sim, "In rec sim this test fails")
+    def test_GIVEN_available_AND_reserved_bytes_set_WHEN_read_THEN_staTus_of_memory_is_as_expected(self):
+        self.ca.set_pv_value("KHLY2700_01:BUFFER_SIZE:SP", 100)
+        self._lewis.backdoor_set_on_device("buffer_readings", 10)
+        self._lewis.backdoor_set_on_device("buffer", 100)
+        self.ca.assert_that_pv_is("KHLY2700_01:BUFFER_SIZE.", 100)
+        self.ca.set_pv_value("KHLY2700_01:BUFFER_STATS.PROC", 1)
+        self.ca.assert_that_pv_is("KHLY2700_01:BUFFER_STATS", "92160 bytes,10240 bytes")
+
+    def test_GIVEN_source_set_WHEN_read_THEN_source_is_as_expected(self):
+        sample_data = [0, 1, 2, 3, 4]
+        expected_channel = ["IMM", "EXT", "TIM", "MAN", "BUS"]
+        for input, output in zip(sample_data, expected_channel):
+            self.ca.set_pv_value("KHLY2700_01:CONTROL_SOURCE:SP", input)
+            self.ca.assert_that_pv_is("KHLY2700_01:CONTROL_SOURCE", output)
+
+    @skipIf(IOCRegister.uses_rec_sim, "In rec sim this test fails")
+    def test_GIVEN_delay_state_set_WHEN_read_THEN_delay_state_is_as_expected(self):
+        sample_data = [0, 1]
+        expected_state = ["OFF", "ON"]
+        for input, output in zip(sample_data, expected_state):
+            self.ca.set_pv_value("KHLY2700_01:DELAY_STATE:SP", input)
+            self.ca.assert_that_pv_is("KHLY2700_01:DELAY_STATE", output)
+
+    def test_GIVEN_buffer_size_set_WHEN_read_THEN_buffer_size_is_as_expected_AND_within_range(self):
+
+        expected_buffer_size = [5500, 0, 2, 70000]
+        alarm_state = [ChannelAccess.ALARM_NONE, ChannelAccess.ALARM_MAJOR, ChannelAccess.ALARM_NONE,
+                            ChannelAccess.ALARM_MAJOR]
+        for input, output in zip(expected_buffer_size, alarm_state):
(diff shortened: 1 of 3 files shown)


tests/keithley_2700.py
```diff
@@ -0,0 +1,188 @@
...
+        self._lewis.backdoor_set_on_device("buffer_readings", 10)
+        self._lewis.backdoor_set_on_device("buffer", 100)
+        self.ca.assert_that_pv_is("KHLY2700_01:BUFFER_SIZE.", 100)
+        self.ca.set_pv_value("KHLY2700_01:BUFFER_STATS.PROC", 1)
+        self.ca.assert_that_pv_is("KHLY2700_01:BUFFER_STATS", "92160 bytes,10240 bytes")
+
+    def test_GIVEN_source_set_WHEN_read_THEN_source_is_as_expected(self):
+        sample_data = [0, 1, 2, 3, 4]
+        expected_channel = ["IMM", "EXT", "TIM", "MAN", "BUS"]
+        for input, output in zip(sample_data, expected_channel):
```
AdrianPotter: Zipping together two lists is risky. As the data isn't in itself coupled, you are open to the possibility of them changing independently and invalidating the tests. You could instead do this as:

```
expected_channels = {0: "IMM", 1: "EXT"...}
for channel_number in expected_channels.keys():
     ...
```

Alternatively, given the form of `sample_data` you could do:

```
for channel_number, channel_name in enumerate(expected_channel):
   ...
```

Please pick an improved pattern and apply it to all other instances where you've used `zip`

tests/keithley_2700.py
```diff
@@ -0,0 +1,188 @@
...
+        self.ca.assert_that_pv_is("KHLY2700_01:IDN", "Keithley 2700 Multimeter emulator.")
+
+    @skipIf(IOCRegister.uses_rec_sim, "In rec sim this test fails")
+    def test_GIVEN_available_AND_reserved_bytes_set_WHEN_read_THEN_staTus_of_memory_is_as_expected(self):
+        self.ca.set_pv_value("KHLY2700_01:BUFFER_SIZE:SP", 100)
+        self._lewis.backdoor_set_on_device("buffer_readings", 10)
+        self._lewis.backdoor_set_on_device("buffer", 100)
+        self.ca.assert_that_pv_is("KHLY2700_01:BUFFER_SIZE.", 100)
+        self.ca.set_pv_value("KHLY2700_01:BUFFER_STATS.PROC", 1)
+        self.ca.assert_that_pv_is("KHLY2700_01:BUFFER_STATS", "92160 bytes,10240 bytes")
```
AdrianPotter: There is a method in `ca` for simultaneously setting a PV and checking the readback matches the expected value. Please use that instead and apply elsewhere in the test file

tests/keithley_2700.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+    Tests for the Keithley2700.
+    """
+
+    def setUp(self):
+        self._lewis, self._ioc = get_running_lewis_and_ioc("keithley_2700")
+        self.ca = ChannelAccess(default_timeout=30)
+        self.ca.wait_for("KHLY2700_01:IDN")
+
+    @skipIf(IOCRegister.uses_rec_sim, "In rec sim this test fails")
```
AdrianPotter: Using "In rec sim this test fails" is a bad habit we've gotten into. Please could we break it. Not running the test in rec sim tells me that it fails in rec sim. I'd like the commenting to tell me what it is about the test that doesn't work.

tests/keithley_2700.py
```diff
@@ -0,0 +1,188 @@
...
+    def test_GIVEN_measurement_mode_set_to_dc_voltage_WHEN_read_then_measurement_mode_is_dc_voltge(self):
+        expected_value = 26.0
+        self._lewis.backdoor_set_on_device("reading", expected_value)
+        expected_configuration = "DCV"
+        self.ca.assert_that_pv_is("KHLY2700_01:MEAS:VOLT:DC", expected_value)
+        device_mode = self._lewis.backdoor_command(["device", "get_channel_param", "1", "mode"],
+                                                   returns_single_value=True)
+        #self.assertEquals(device_mode, expected_configuration)
```
AdrianPotter: Why is this commented out?

tests/keithley_2700.py
```diff
@@ -0,0 +1,188 @@
...
+        for input, output in zip(sample_data, expected_state):
+            self.ca.set_pv_value("KHLY2700_01:STAMP:SP", input)
+            self.ca.assert_that_pv_is("KHLY2700_01:STAMP", output)
+
+    @skipIf(IOCRegister.uses_rec_sim, "In rec sim this test fails")
+    def test_GIVEN_measurement_resolution_set_WHEN_read_THEN_measurement_resolution_is_as_expected(self):
+        self.ca.set_pv_value("KHLY2700_01:CHANNEL_END:SP", 110)
+        self.ca.set_pv_value("KHLY2700_01:DIGIT_RANGE:SP", 6)
+        self.ca.assert_that_pv_is("KHLY2700_01:DIGIT_RANGE", 6)
+        self.ca.assert_that_pv_is("KHLY2700_01:CHANNEL_END", 110)
```
AdrianPotter: If you use a value more than once, it's best practice to put it in a variable so that if we change it in one place, it changes everywhere

tests/keithley_2700.py
```diff
@@ -0,0 +1,188 @@
...
+
+    @skipIf(IOCRegister.uses_rec_sim, "In rec sim this test fails")
+    def test_GIVEN_data_string_set_WHEN_read_THEN_data_string_is_as_expected(self):
+        self.ca.set_pv_value("KHLY2700_01:LIST_OF_ELEMENTS:SP", "READ,CHAN,TST")
+        self.ca.assert_that_pv_is("KHLY2700_01:LIST_OF_ELEMENTS:SP", "READ,CHAN,TST")
+        self._lewis.backdoor_set_on_device("reading", 45.0)
+        self._lewis.backdoor_set_on_device("channel_number", 110)
+        self._lewis.backdoor_set_on_device("time_stamp", 4)
+        self.ca.set_pv_value("KHLY2700_01:LIST_OF_ELEMENTS.PROC", 1)
+        self.ca.assert_that_pv_is("KHLY2700_01:LIST_OF_ELEMENTS", "45.0,110,4")
```
AdrianPotter: The expected string is constructed from the input values. As previously, it's better to not hard code it. Extract the input values into variables and use those variables to construct the expected return string. It makes the test less fragile and gives a much better impression of how the input corresponds to the output.

tests/lakeshore460.py
```diff
@@ -0,0 +1,202 @@
...
+    @skipIf(IOCRegister.uses_rec_sim, "In rec sim this test fails")
+    def test_GIVEN_magnetic_field_reading_unit_set_WHEN_read_THEN_magnetic_field_reading_unit_is_as_expected(self):
+        sample_data = [0, 1, 2, 3]
+        expected_unit = ["uG", "mG", "G", "kG"]
+        for input, output in zip(sample_data, expected_unit):
+            self.ca.set_pv_value("LKSH460_01:FIELD:MULTIPLIER", input)
+            self.ca.assert_that_pv_is("LKSH460_01:FIELDREADING.EGU", output)
+
+
+
```
AdrianPotter: The Lakeshore tests shouldn't be on this branch.

AdrianPotter:
https://github.com/ISISComputingGroup/IBEX/issues/647

PR #50, opened 2018-01-17
Ticket2813 riken changeover logic

Tom-Willemsen:
This change to the IOC test framework was required for the riken FE changeover logic (https://github.com/ISISComputingGroup/IBEX/issues/2813)

The test framework was previously unable to launch multiple IOCs in one test. I've done a fairly large refactor to make the test suites "know" what IOCs to launch as opposed to this being passed in from batch files.

As a side effect of this, I was able to get rid of the ever-growing list of calls to the test framework in the run_all_tests batch file. Python now automatically detects what test modules are available, runs them in recsim and/or devsim as appropriate, and collects all the results. This last refactoring wasn't strictly required for this ticket but I felt it was a worthwhile improvement to make while I was refactoring he test framework anyway.

Reviewer: because this is a substantial change, please could you let me know when this is ready to be merged so that I can update the documentation.


run_tests.py
```diff
@@ -53,117 +59,185 @@ def modified_environment(**kwargs):
...
     # Define an environment variable with the prefix in it
     # This can then be accessed elsewhere
     os.environ["testing_prefix"] = prefix
 
-    # Load the device's test module and get the class
-    m = load_module('tests.%s' % device.lower())
+    # Need to set epics address list to local broadcast otherwise channel access won't work
+    settings = {
+        'EPICS_CA_ADDR_LIST': "127.255.255.255"
+    }
```
DominicOram: Why can't you use the address list that comes from config_env? I assume because you don't want the test framework looking at the gateway? If so, please add a comment to this effect.

Tom-Willemsen: This isn't behaviour I've changed - I'm not quite sure why it was done in that way in the first place. I imagine you're correct about the gateway.

Github hasn't given the most helpful diff but see old version of `run_tests.py`, line 91 for where this was being done before.

DominicOram: Apologies, we'll keep as is for now then.

run_tests.py
```diff
@@ -53,117 +59,185 @@ def modified_environment(**kwargs):
...
+        print("Cannot run without instrument prefix")
+        sys.exit(-1)
+
+    if arguments.emulator_path is None:
+        print("Cannot run without emulator path")
+        sys.exit(-1)
+
+    module_results = []
+
+    if arguments.test_module is not None:
```
DominicOram: This is a bit copy-and-paste can you join the test_module and package_contents lists then loop through them together?

run_tests.py
```diff
@@ -53,117 +59,185 @@ def modified_environment(**kwargs):
...
-        ioc_launcher.device_prefix = m.DEVICE_PREFIX
+        modes = test_module.TEST_MODES
     except AttributeError:
-        ioc_launcher.device_prefix = None
+        raise ValueError("Expected test module {} to contain a TEST_MODES attribute".format(test_module.__name__))
 
-    # Override any emulator port that might be set elsewhere
-    ioc_launcher.macros['EMULATOR_PORT'] = port
+    test_results = []
+    for mode in modes:
```
DominicOram: This means that if a list of N elements then it will run the test in devsim N times, regardless of what those elements are. Maybe worth some checks that the list is a set and only contains valid TestModes?

run_tests.py
```diff
@@ -53,117 +59,185 @@ def modified_environment(**kwargs):
...
+                                   directory=ioc["directory"],
+                                   macros=macros,
+                                   use_rec_sim=recsim,
+                                   var_dir=var_dir,
+                                   port=free_port)
+
+        if "emulator" in ioc and not recsim:
+
+            emulator_name = ioc["emulator"]
+            try:
```
DominicOram: Can you use get with a default for these few lines?

utils/ioc_launcher.py
```diff
@@ -1,10 +1,27 @@
 import subprocess
 import os
+from time import sleep
 
 from utils.channel_access import ChannelAccess
 from utils.log_file import log_filename
 
 
+EPICS_TOP = os.path.join("C:\\", "Instrument", "Apps", "EPICS")
```
DominicOram: This is set in an environment variable called KIT_ROOT when running config_env. Might be better to use that?

run_tests.py
```diff
@@ -53,117 +59,185 @@ def modified_environment(**kwargs):
...
+    print("Testing module {} in {} mode.".format(test_module.__name__, "recsim" if recsim else "devsim"))
+
+    device_launchers = []
+    for ioc in iocs:
+
+        free_port = str(get_free_ports(1)[0])
+        macros = ioc["macros"]
+        macros['EMULATOR_PORT'] = free_port
+
+        ioc_launcher = IocLauncher(device=ioc["name"],
```
DominicOram: Can you put in a better error message for when you don't specify the required keys. Just getting an `KeyError: name` for the test you've just created isn't very obvious.

run_tests.py
```diff
@@ -53,117 +59,185 @@ def modified_environment(**kwargs):
...
+    if len(iocs) < 1:
+        raise ValueError("Need at least one IOC to launch")
+
+    print("Testing module {} in {} mode.".format(test_module.__name__, "recsim" if recsim else "devsim"))
+
+    device_launchers = []
+    for ioc in iocs:
+
+        free_port = str(get_free_ports(1)[0])
+        macros = ioc["macros"]
```
DominicOram: Is it worth putting a default of an empty dict for the macros?

DominicOram:
I think it's much cleaner than it was. I may have been a bit pedantic on some of my comments, apologies.

common_tests/riken_changeover.py
```diff
@@ -0,0 +1,141 @@
...
+        self.ca.wait_for("{}:PSUS:DISABLE".format(self.get_prefix()), timeout=30)
+        self.ca.wait_for(self.get_input_pv(), timeout=30)
+        self.ca.wait_for(self.get_acknowledgement_pv(), timeout=30)
+        for id in self.get_power_supplies():
+            self.ca.wait_for("{}:POWER".format(id), timeout=30)
+
+        self._set_input_pv(True)
+        self._set_all_power_supply_states(False)
+
+    def test_GIVEN_value_on_input_ioc_changes_THEN_coord_picks_up_the_change(self):
```
DominicOram: This name is a bit vague.

common_tests/riken_changeover.py
```diff
@@ -0,0 +1,141 @@
...
+    def test_GIVEN_plc_cancels_changeover_before_psus_are_all_switched_off_WHEN_psus_become_switched_off_THEN_they_do_not_get_disabled(self):
+        self._set_all_power_supply_states(True)
+        self._set_input_pv(False)
+        self._assert_all_power_supplies_disabled(False)  # Power supplies not disabled because still powered on
+        self._set_input_pv(True)  # PLC now cancels request to do a changeover
+        self._set_all_power_supply_states(False)
+        self._assert_all_power_supplies_disabled(False)
+
+    # Using a context manager to put PVs into alarm means they don't accidentally get left in alarm if the test fails
+    @contextmanager
```
DominicOram: Useful context manager, is it worth putting into the IocTestFramework core?

common_tests/riken_changeover.py
```diff
@@ -0,0 +1,141 @@
...
+            self.ca.set_pv_value("{}:POWER.SIMS".format(supply), alarm)
+            self.ca.assert_pv_alarm_is("{}:POWER".format(supply), alarm)
+
+        try:
+            _set_and_check_simulated_alarm(supply, self.ca.ALARM_INVALID)
+            yield
+        finally:
+            _set_and_check_simulated_alarm(supply, self.ca.ALARM_NONE)
+
+    def test_GIVEN_a_power_supply_is_in_alarm_THEN_the_power_any_pv_is_also_in_alarm(self):
```
DominicOram: For completeness can you test this is also the case when all are in alarm? Same for function below it.

common_tests/riken_changeover.py
```diff
@@ -0,0 +1,141 @@
...
+                self.ca.assert_pv_alarm_is("{}:PSUS:POWER".format(self.get_prefix()), self.ca.ALARM_INVALID)
+            self.ca.assert_pv_alarm_is("{}:PSUS:POWER".format(self.get_prefix()), self.ca.ALARM_NONE)
+
+    def test_GIVEN_a_power_supply_is_in_alarm_THEN_the_power_any_pv_reports_that_psus_are_active(self):
+        for supply in self.get_power_supplies():
+            with self._put_power_supply_into_alarm(supply):
+                self.ca.assert_that_pv_is_number("{}:PSUS:POWER".format(self.get_prefix()), 1)
+            self.ca.assert_that_pv_is_number("{}:PSUS:POWER".format(self.get_prefix()), 0)
+
+    def test_GIVEN_changeover_initiated_WHEN_power_supplies_off_THEN_acknowledgement_pv_true(self):
```
DominicOram: Can you add a test for the PSUs to be re-enabled after changeover?

DominicOram:
There's some more code in the two test files that could probably be moved to the common file:
* The building of the IOC list could be a common function
* The building of the power supplies list can be done in the class

Tom-Willemsen:
documentation updated

PR #53, opened 2018-01-30
Ticket2828 add ioc triton

Tom-Willemsen:
IOC system tests for https://github.com/ISISComputingGroup/IBEX/issues/2828


utils/channel_access.py
```diff
@@ -434,3 +437,24 @@ def assert_pv_value_is_unchanged(self, pv, wait):
         :raises AssertionError: if the value of the pv has changed
         """
         self.assert_pv_value_over_time(pv, wait, operator.eq)
+
+    def assert_pv_value_causes_func_to_return_true(self, pv, func, timeout=None):
```
AdrianPotter: I wonder if this would be better as:

```
def assert_function_applied_to_pv_value_matches_return(self, pv, func, return_value, timeout=None):
    ...
```

The function gives total flexibility over input functions but fixes it on a return value of `True`. I would imagine flexibility on the return value might be more valuable in the long run.

Tom-Willemsen: To be honest I think I prefer the form
`assert_function_returns_true(lambda val: my_complex_function(val) == 47)`
over
`assert_function_returns_value(lambda val: my_complex_function(val), 47)`

I agree that the general API of the assertions in the test framework is getting pretty messy, but making significant changes there is well beyond the scope of this ticket

PR #55, opened 2018-01-31
Ticket 2138: GEM beamscrapper jaws

DominicOram:
To test see: https://github.com/ISISComputingGroup/EPICS-motorExtensions/pull/10


tests/gem_jaws.py
```diff
@@ -0,0 +1,89 @@
...
+        # Uses a sim record to save the time for the simulated motor to stop
+        self.ca.set_pv_value(calibrated_axis + ":SIM:MTR:RBV", x)
+        self.ca.assert_that_pv_is_number(calibrated_axis + ".RBV", to_read_func(x), TOLERANCE)
+
+    def _test_set_point(self, underlying_motor, calibrated_axis, to_write_func, x):
+        self.ca.set_pv_value(calibrated_axis, x)
+        self.ca.assert_that_pv_is_number(underlying_motor + ".VAL", to_write_func(x), TOLERANCE)
+
+    # Quadratic axis tests
+
```
davidkeymer: It's been suggested that the expected values for the following three tests be refactored to variables to help explain why they've been chosen.  e.g. MIN_POSITION = 2, MID_POSITION = 5.4, MAX_POSITION = 12.  It would also clarify why they're expected in the first place.

PR #57, opened 2018-02-05
Fix tests

Tom-Willemsen:


Tom-Willemsen:
Flash review by @John-Holt-Tessella

PR #85, opened 2018-04-06
Ticket 3086: Added tests for limits on gem beamscrapers

DominicOram:
Added test for limits and simplified some of the other tests.

I have removed the stopping test as it was previously passing with a false positive. I spent some time trying to fix it but the IOC Test Framework cannot seem to stop simulated motors (see attached test). If a reviewer agrees that this doesn't work I'll add a new ticket.
[motor_sim.zip](https://github.com/ISISComputingGroup/EPICS-IOC_Test_Framework/files/1885395/motor_sim.zip)


DominicOram:
Tests appear to work for @Tom-Willemsen, I will add them back in and try and work out why they're broken locally at a later date.

PR #93, opened 2018-05-23
Modify IOC stopping logic

Tom-Willemsen:
Silence CA errors, and use better logic to tell if an IOC is dead

https://github.com/ISISComputingGroup/IBEX/issues/3195


DominicOram:
CA errors are no longer outputted and tests seem more stable

PR #95, opened 2018-05-30
Ticket3214 ioc tpg300

Rory-Potter-Tessella:
### Issue

[#3214](https://github.com/ISISComputingGroup/IBEX/issues/3214).

### Description of Work Done

Adds tests for an emulator of TPG300 using DEVSIM and using the SIM record files using RECSIM.

### Tests

- IOC is not disabled on start
- The correct units are set on PV using SIM and emulator. 
- Postive floats are set correctly.
- Negative floats are set correctly.
- Integers are set correctly.
- Exponential values are set correctly.
- Alarm flags when the emulator is disconnected.

### DIFF
--- a/tests/tpg300.py
+++ b/tests/tpg300.py
@@ -35,9 +35,11 @@ class Tpg300Tests(unittest.TestCase):
 
         self.ca = ChannelAccess(20, device_prefix=DEVICE_PREFIX)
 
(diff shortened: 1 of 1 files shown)


Rory-Potter-Tessella:
Updates since review:

- Condensed pressure tests into one test as suggested.
- Units are now an enum class.

PR #104, opened 2018-06-20
Refactor assertions

Tom-Willemsen:
https://github.com/ISISComputingGroup/IBEX/issues/3281

Refactor the `ChannelAccess` class in the test framework.

The following has been done:
- Consolidate all functions to call a common one, `assert_that_pv_value_causes_func_to_return_true`. This removes a lot of duplicate logic
- Refactor the alarms to be a class within `ChannelAccess` rather than constants with a common prefix
- Convert documentation to new style
- Adjust assertion naming to improve consistency


DominicOram:
Looks good, I refactored the cybaman a little to use better asserts.

PR #117, opened 2018-08-09
modified test

LiamPanchaud:
Included tolerances


Tom-Willemsen:
I think you should be able to add a test that checks that the ramp units are set. I believe all the infrastructure to change calibration files etc is already in place.

LiamPanchaud:
You were right, I hadn't checked because someone said it would be a nightmare to implement. Presumably they didn't know about the infrastructure already there.

PR #121, opened 2018-08-28
Ticket3071 add ioc kicker

Rory-Potter-Tessella:
### Description of work done

- Added RECSIM tests to test voltage and current records for communicating with the DAQ.
- Updated Readme to explain how to include a suite of tests as a module.

### Ticket

[#3404](https://github.com/ISISComputingGroup/IBEX/issues/3404)


tests/kicker/test_basic_commands.py
```diff
@@ -0,0 +1,81 @@
...
+        self.ca.set_pv_value("{}:SIM".format(pv_root), value)
+        self.ca.assert_that_pv_is("{}:SIM".format(pv_root), value)
+
+        value_to_check = self.ca.get_pv_value("{}:_RAW".format(pv_root))[0]
+        assert_that(value_to_check, is_(equal_to(value)))
+
+    @parameterized.expand(
+        parameterized_list([4.68, 6, 10, 0, 4e-3])
+    )
+    def test_that_GIVEN_a_value_THEN_the_calibrated_value_is_read(self, _, value_to_set):
```
Alistair-McGann-Tessella: The names of these tests are very vague, which I think is because of the multiple purposes each test needs to conform to as part of the test suite. As a result the meaning of each test is hidden and can make them hard to read, and it is more difficult to diagnose a failed test. Similar for other tests.

tests/kicker/test_basic_commands.py
```diff
@@ -0,0 +1,81 @@
...
+    record = None
+    calibration = None
+
+    def setUp(self):
+        self.ca = ChannelAccess(20, device_prefix=DEVICE_PREFIX)
+        self.ca.assert_that_pv_is("DISABLE", "COMMS ENABLED")
+        self._simulate_value(0)
+
+    def _simulate_value(self, value):
+        pv_root = "DAQ:{}".format(self.record)
```
Alistair-McGann-Tessella: This method depends on the state that the class is in, which makes reading/debugging challenging. Could the data held in `self.record` be passed in as an argument instead?

tests/kicker/test_basic_commands.py
```diff
@@ -0,0 +1,81 @@
...
+    )
+    def test_that_GIVEN_an_edge_value_THEN_the_pv_is_in_alarm(self, _, value_to_set):
+        # Given:
+        self._simulate_value(value_to_set)
+        expected_calibrated_value = self.calibration * value_to_set
+        self.ca.assert_that_pv_is_number(self.record, expected_calibrated_value, 0.01)
+
+        # Then:
+        self.ca.assert_that_pv_alarm_is(self.record, self.ca.Alarms.MAJOR)
+
```
Alistair-McGann-Tessella: If the waveform records cannot be tested, could there be a comment somewhere explaining why? Similar for the Schneider PLC, but I understand that can't be interfaced with yet.

tests/kicker/test_basic_commands.py
```diff
@@ -0,0 +1,81 @@
...
+    def test_that_GIVEN_a_value_THEN_the_calibrated_value_is_read(self, _, value_to_set):
+        # Given:
+        self._simulate_value(value_to_set)
+
+        # Then:
+        expected_calibrated_value = self.calibration * value_to_set
+        self.ca.assert_that_pv_is_number(self.record, expected_calibrated_value, 0.01)
+
+    @parameterized.expand(
+        parameterized_list([15, -5])
```
Alistair-McGann-Tessella: Are the current and voltage records tested using the same values? This might have unintended consequences when testing against current/voltage limits.

PR #126, opened 2018-09-24
Ticket3590 abort after failed tests

LiamPanchaud:
[Ticket 3590](https://github.com/ISISComputingGroup/IBEX/issues/3590)

PR #142, opened 2018-12-18
Ticket2994 keithley 2700 cryomag tests

LiamPanchaud:
NOTE: I accidentally based this from the wrong branch, I believe is has code from the separator and a friday ticket (#126) present. I think the separator has been merged, but the friday ticket hasn't at the time of writing. 

(Also, the tests aren't written too well, there's a fair amount of code reuse. This is still being addressed as the Keithley 2700 stuff is still in development. Still, of course, all feedback on improvements rework would be greatly appreciated)


tests/keithley_2700.py
```diff
@@ -16,17 +21,23 @@
     },
 ]
 
-TEST_MODES = [TestModes.RECSIM, TestModes.DEVSIM]
```
aaron-long: Looks like you forgot to restore the full suite of test modes.

tests/keithley_2700.py
```diff
@@ -150,13 +153,241 @@ def test_WHEN_end_channel_range_set_THEN_end_channel_matches_the_set_state(self)
...
+        self.ca = ChannelAccess(default_timeout=30, device_prefix=DEVICE_PREFIX)
+        self.ca.assert_that_pv_exists("IDN")
+        self.ca.set_pv_value("BUFF:CLEAR:SP", "")
+        self.ca.assert_that_pv_is("BUFF:AUTOCLEAR", "ON")
+
+    def _generate_readings(self, num_readings_gen, time_between):
+        min_r = 1000
+        max_r = 2000
+        start_time = 1
+        channels = [c for c in range(101, 110 + 1)] + [c for c in range(201, 210 + 1)]
```
aaron-long: What do these channel numbers mean? Maybe leave a comment to explain, or pull them out like min_r/max_r

tests/keithley_2700.py
```diff
@@ -150,13 +153,241 @@ def test_WHEN_end_channel_range_set_THEN_end_channel_matches_the_set_state(self)
...
+        readings = []
+
+        for i in range(num_readings_gen):
+            resistance = min_r + i
+            while resistance > max_r:
+                resistance = resistance - min_r
+            time_stamp = start_time + (i*time_between)
+            channel = channels[i % 20]
```
aaron-long: Again, like the previous channels comment; this is slightly obscure as to what it is doing.

tests/keithley_2700.py
```diff
@@ -150,13 +153,241 @@ def test_WHEN_end_channel_range_set_THEN_end_channel_matches_the_set_state(self)
...
+            readings.append(str(resistance) + "," + str(time_stamp) + "," + str(channel))
+
+        return readings
+
+    def _set_buffer_size(self, buff_size):
+        self.ca.set_pv_value("BUFF:SIZE:SP", buff_size)
+        self.ca.assert_that_pv_is("BUFF:SIZE", buff_size)
+
+    @contextmanager
```
aaron-long: Context manager might be overkill because it does nothing on the `finally`, it can just be replaced by a function - Oram, Dom (2019)

tests/keithley_2700.py
```diff
@@ -150,13 +153,241 @@ def test_WHEN_end_channel_range_set_THEN_end_channel_matches_the_set_state(self)
...
         self.ca.assert_setting_setpoint_sets_readback(elements, "DATAELEMENTS")
+
+
+class BufferTests(unittest.TestCase):
+    def setUp(self):
+        self._lewis, self._ioc = get_running_lewis_and_ioc("keithley_2700", DEVICE_PREFIX)
+        self.ca = ChannelAccess(default_timeout=30, device_prefix=DEVICE_PREFIX)
+        self.ca.assert_that_pv_exists("IDN")
+        self.ca.set_pv_value("BUFF:CLEAR:SP", "")
+        self.ca.assert_that_pv_is("BUFF:AUTOCLEAR", "ON")
```
aaron-long: There doesn't seem to be tests for the case of `BUFF:AUTOCLEAR, OFF`. If needed then make some tests.

tests/keithley_2700.py
```diff
@@ -150,13 +153,241 @@ def test_WHEN_end_channel_range_set_THEN_end_channel_matches_the_set_state(self)
...
+        # insert final reading
+        with self._insert_reading([reads[5]]):
+            self.ca.assert_that_pv_is("BUFF:NEXT", 0)
+            self.ca.assert_that_pv_is("INDEX:START", 0)
+
+    @skip_if_recsim("Cannot use lewis backdoor in recsim")
+    def test_GIVEN_buffer_full_WHEN_new_reading_added_THEN_new_reading_added_at_index_0_AND_buffer_still_used(self):
+        buffer_size = 10
+        self._set_buffer_size(buffer_size)
+        reads = self._generate_readings(15, 5)  # 10 greater than buffer capacity
```
aaron-long: Comment is incorrect & unneeded. Dom

tests/keithley_2700.py
```diff
@@ -150,13 +153,241 @@ def test_WHEN_end_channel_range_set_THEN_end_channel_matches_the_set_state(self)
...
+    def reset_drift_channels(self):
+        channel_nums = [c for c in range(101, 110 + 1)] + [c for c in range(201, 210 + 1)]
+        channels = ["CHNL:" + str(c) for c in channel_nums]
+
+        channel_drifts = [c + ":DRIFT" for c in channels]
+
+        for pv in channel_drifts:
+            self.ca.set_pv_value(pv, 0)
+
+    @contextmanager
```
aaron-long: Again, the context manger is no really required.

tests/keithley_2700.py
```diff
@@ -150,13 +153,241 @@ def test_WHEN_end_channel_range_set_THEN_end_channel_matches_the_set_state(self)
...
+    def reset_drift_channels(self):
+        channel_nums = [c for c in range(101, 110 + 1)] + [c for c in range(201, 210 + 1)]
+        channels = ["CHNL:" + str(c) for c in channel_nums]
+
+        channel_drifts = [c + ":DRIFT" for c in channels]
+
+        for pv in channel_drifts:
+            self.ca.set_pv_value(pv, 0)
+
+    def _lewis_sync_helper(self, attribute, set_value, wait_time=0.2):
```
aaron-long: Remove if unused

tests/keithley_2700.py
```diff
@@ -16,17 +21,23 @@
     },
 ]
 
-TEST_MODES = [TestModes.RECSIM, TestModes.DEVSIM]
+# TEST_MODES = [TestModes.RECSIM, TestModes.DEVSIM]
+TEST_MODES = [TestModes.DEVSIM]
 
 on_off_status = {False: "OFF", True: "ON"}
 
+DRIFT_TOLERANCE = 0.1
```
aaron-long: Might be appropriate to have a tolerance on the same order of magnitude as the drift values.

tests/keithley_2700.py
```diff
@@ -150,13 +153,241 @@ def test_WHEN_end_channel_range_set_THEN_end_channel_matches_the_set_state(self)
...
+        channel_drifts = [c + ":DRIFT" for c in channels]
+
+        for pv in channel_drifts:
+            self.ca.set_pv_value(pv, 0)
+
+    def _lewis_sync_helper(self, attribute, set_value, wait_time=0.2):
+        while self._lewis.backdoor_get_from_device(str(attribute)) != str(set_value):
+            time.sleep(wait_time)
+
+    @contextmanager
```
aaron-long: Same comment about the others.

tests/keithley_2700.py
```diff
@@ -150,13 +153,241 @@ def test_WHEN_end_channel_range_set_THEN_end_channel_matches_the_set_state(self)
...
+        reads = self._generate_readings(15, 5)
+        expected_reads = reads[7:10]
+
+        with self._insert_reading(reads[:7]):
+            self.ca.assert_that_pv_is("BUFF:NEXT", 7)
+
+        with self._insert_reading(reads[7:10]):  # This fills the buffer
+            self.ca.assert_that_pv_is("BUFF:NEXT", 0)
+
+        retrieved_readings = self.ca.get_pv_value("BUFF:READ")[:9]
```
aaron-long: For the `BUFF:READ` calls, I think it would be useful to have a comment to clarify that you'll return 3 values for each inserted reading.

LiamPanchaud:
Forgot to do one final push in which context managers were removed and functions extracted from classes etc. 

I'll push it along with the changes

tests/keithley_2700.py
LiamPanchaud: The device should never operate with autoclear off in its current use case

tests/keithley_2700.py
LiamPanchaud: Needs refining once the new interpolation routine is in place

PR #150, opened 2019-01-14
Ticket 3929 - MOXA e1210: Added test that checks the counter correctly increases.

aaron-long:
Each DI can be triggered which will increament the DI's associated counter (EPICS layer record). Typically we want these counter values to remain persistent even on restart (as they are a safety feature). 

The new tests trigger each DI and then check their counter has incremented correctly. The resetDICounter function allows for these to be reset, but add considerable time to the test length, as such it shouldn't be in setUP.

### DIFF
--- a/tests/moxa1210.py
+++ b/tests/moxa1210.py
@@ -1,5 +1,6 @@
 from __future__ import division
 import unittest
+from random import randint
 from time import sleep
 
 from utils.test_modes import TestModes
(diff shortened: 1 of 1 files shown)


tests/moxa1210.py
```diff
@@ -55,3 +66,12 @@ def test_WHEN_DI_input_is_switched_on_THEN_only_that_channel_readback_changes_to
                 continue
 
             self.ca.assert_that_pv_is("CH{:02d}:DI".format(test_channel), "Low")
+
+    @parameterized.expand([
+        ("CH{:02d}:DI:CNT".format(channel), channel) for channel in CHANNELS
+    ])
+    def test_WHEN_di_input_is_triggered_THEN_di_counter_increases(self, channel_pv, channel):
+        self.resetDICounters()
```
Alistair-McGann-Tessella: This function appears to reset all of the counters when testing one channel. Might it speed the running of the tests up by only resetting each channel as it's being tested?

tests/moxa1210.py
```diff
@@ -55,3 +66,12 @@ def test_WHEN_DI_input_is_switched_on_THEN_only_that_channel_readback_changes_to
...
+        self._lewis.backdoor_run_function_on_device("set_di", (channel, (True,)))
+
+        self.ca.assert_that_pv_is(channel_pv, 1)
```
Alistair-McGann-Tessella: Would it be useful to demonstrate that the counters are properly adding up over multiple switches by turning a DI on and off again a few times?

tests/moxa1210.py
aaron-long: I changed the reset function to update only the channel the test applies to.

tests/moxa1210.py
aaron-long: Created a new test to check that individual channels would correctly log multiple events.

PR #170, opened 2019-03-21
Fix PV names

FreddieAkeroyd:
Fix PV names

See ISISComputingGroup/IBEX#4161


Tom-Willemsen:
We should add tests for the second shutter - I added sub-ticket I added a sub-ticket https://github.com/ISISComputingGroup/IBEX/issues/4194 to do this.

PR #175, opened 2019-04-04
Fix backdoor run on device

Tom-Willemsen:


Tom-Willemsen:
Flash reviewed by @John-Holt-Tessella

PR #178, opened 2019-04-15
Ticket3542 refl init setpoints

ThomasLohnert:
Added IOC tests for https://github.com/ISISComputingGroup/IBEX/issues/3542

In order to make them work, I added the following functionality to the ioc launcher:
- You can initialise PV values of an IOC on startup by passing into the IOC dict an item `inits`, which is itself a dict of the form `PV`: `init value`.
- You can specify a custom `PV` prefix by specifying a `custom_prefix` item in the IOC dict. (e.g. `MOT` for IOC `GALIL_01`)

Related PR: https://github.com/ISISComputingGroup/EPICS-inst_servers/pull/198


utils/ioc_launcher.py
```diff
@@ -428,7 +433,8 @@ def _get_channel_access(self):
         :return (ChannelAccess): the channel access component
         """
         if self._ca is None:
-            self._ca = ChannelAccess(device_prefix=self._device)
+            prefix = self._custom_prefix or self._device
```
John-Holt-Tessella: I learn more python everyday.

PR #189, opened 2019-05-23
Add tests for explicitly setting calc mode

ThomasLohnert:
Added tests to check the behaviour of the IOC is correct when selecting either of the two calculation methods for running the pump in timed mode. 

I.e. writing values to `TIME:RUN:SP` and subsequent calculation of `TIME:VOL:SP` is only allowed when `TIME:CALC:SP` is set to `Time`, and vice versa for `Volume`.

Ticket:
https://github.com/ISISComputingGroup/IBEX/issues/4029

Related PRs:
https://github.com/ISISComputingGroup/ibex_gui/pull/891
https://github.com/ISISComputingGroup/EPICS-Jsco4180/pull/4

### DIFF
--- a/tests/jsco4180.py
+++ b/tests/jsco4180.py
@@ -37,6 +37,7 @@ class Jsco4180Tests(unittest.TestCase):
 
     def resetIocRecords(self):
         self.ca.set_pv_value("TIME", 0)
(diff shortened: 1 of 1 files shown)


SophieKSTFC:
Looks good, tests work as expected.

PR #204, opened 2019-07-23
Added simple tests to confirm split jaws records work correctly

ThomasLohnert:
Added tests for https://github.com/ISISComputingGroup/EPICS-jaws/pull/8

### DIFF
--- a/test_config/motor_test/jaws.cmd
+++ b/test_config/motor_test/jaws.cmd
@@ -1,3 +1,0 @@
-$(IFIOC_GALIL_01=#) dbLoadRecords("$(JAWS)/db/split_jaws_header_vertical.db","P=$(MYPVPREFIX)MOT:,JAWS=JAWS1:,mXN=MTR0101,mXS=MTR0102,mXW=MTR0202,mXE=MTR0201")
-
-$(IFIOC_GALIL_02=#) dbLoadRecords("$(JAWS)/db/split_jaws_horizontal.db","P=$(MYPVPREFIX)MOT:,JAWS=JAWS1:,mXW=MTR0202,mXE=MTR0201")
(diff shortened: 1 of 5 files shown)


test_config/motor_test/jaws.cmd
```diff
@@ -0,0 +1,3 @@
+$(IFIOC_GALIL_01=#) dbLoadRecords("$(JAWS)/db/split_jaws_header_vertical.db","P=$(MYPVPREFIX)MOT:,JAWS=JAWS1:,mXN=MTR0101,mXS=MTR0102,mXW=MTR0202,mXE=MTR0201")
```
DominicOram: If you look at how we test the `motorExtensions` we put the test settings files into the `motorExtensions` directory, I think putting this in the `jaws` directory might be better.

tests/split_jaws.py
```diff
@@ -0,0 +1,129 @@
...
+    Tests for vertical jaws
+    """
+    def setUp(self):
+        self._ioc = IOCRegister.get_running("jaws")
+        self.ca = ChannelAccess(default_timeout=30)
+        for axis in UNDERLYING_AXES:
+            self.ca.set_pv_value("{}.VMAX".format(axis), 100)
+            self.ca.set_pv_value("{}.VELO".format(axis), 100)
+
+    def test_GIVEN_ioc_started_THEN_north_underlying_mtr_fields_can_be_read(self):
```
DominicOram: The first four tests can easily be paramterized.

test_config/motor_test/jaws.cmd
ThomasLohnert: Done. Added folders for each type of jawset in `jaws\settings\` containing a `.cmd` each and updated the path in `vertical_jaws.py` IOC tests to match.

tests/split_jaws.py
ThomasLohnert: With the latest changes, there are 2 test classes which both run the same set of tests for regular and split jaws respectively. With parameterized tests, I could not overwrite the underlying motor PV addresses in the split jaws test class, so I have opted against doing this.

ThomasLohnert:
- Refactored tests so that the same set is run against jawsets with different setups of their underlying motors. 
- Added tests for ABLE and LOCK status.
- Could not reproduce the failure, let me know if you still see it and we can investigate.

DominicOram:
Thanks for the refactor and the additional tests.

I fixed the issue on my machine by setting the gaps to zero on setUp, the underlying issue is https://github.com/ISISComputingGroup/IBEX/issues/4645

PR #211, opened 2019-07-30
Ticket 4611: Double check existence PV and remove autosave files

DominicOram:
See https://github.com/ISISComputingGroup/IBEX/issues/4611

This PR does a few things:
* Hardens up the ATTOCUBE and JAW MANAGER tests
* Removes test autosave files between tests (see https://github.com/ISISComputingGroup/EPICS/pull/180)
* Warns if the PV that is being checked to confirm the IOC is not running still doesn't exist after running the IOC
* Bit of refactoring of the ioc_launcher to take advantage of OOP

To test:
* Confirm ATTOCUBE and JAWS_MANAGER tests work
* Confirm autosave files are removed before each test
* Run a test that checks for a PV that doesn't exist ever and confirm that it warns after the IOC has been run

### DIFF
--- a/utils/ioc_launcher.py
+++ b/utils/ioc_launcher.py
@@ -39,22 +39,31 @@ def get_default_ioc_dir(iocname, iocnum=1):
     return os.path.join(EPICS_TOP, "ioc", "master", iocname, "iocBoot", "ioc{}-IOC-{:02d}".format(iocname, iocnum))
 
 
-def check_if_ioc_already_running(ca, device, test_pv="DISABLE"):
+class check_existence_pv:
     """
-    Checks to see if a IOC is already running by asserting that a pv does not exist.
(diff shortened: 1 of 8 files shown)


utils/ioc_launcher.py
```diff
@@ -39,22 +39,31 @@ def get_default_ioc_dir(iocname, iocnum=1):
     return os.path.join(EPICS_TOP, "ioc", "master", iocname, "iocBoot", "ioc{}-IOC-{:02d}".format(iocname, iocnum))
 
 
-def check_if_ioc_already_running(ca, device, test_pv="DISABLE"):
+class check_existence_pv:
```
Tom-Willemsen:
```suggestion
class check_existence_pv(object):
```

PR #225, opened 2019-08-28
Add test for jaw centre not being changed when setting size at modera…

Tom-Willemsen:
Add a test verifying that setting the moderator/sample size does not change the existing centres of each jaw set (this requirement is from NIMROD).

See https://github.com/ISISComputingGroup/IBEX/issues/4672

### DIFF
--- a/common_tests/jaws_manager_utils.py
+++ b/common_tests/jaws_manager_utils.py
@@ -37,6 +37,21 @@ class JawsManagerBase(object):
             self.ca.assert_that_pv_is_number(UNDERLYING_CENT_SP.format(jaw, direction), 10, 0.1)
             self.ca.assert_that_pv_is_number(UNDERLYING_GAP_SP.format(jaw, direction), expected_gaps[jaw - 1], 0.1)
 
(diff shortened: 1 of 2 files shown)


common_tests/jaws_manager_utils.py
```diff
@@ -37,6 +37,21 @@ def _test_WHEN_centre_is_changed_THEN_centres_of_all_jaws_follow_and_gaps_unchan
             self.ca.assert_that_pv_is_number(UNDERLYING_CENT_SP.format(jaw, direction), 10, 0.1)
             self.ca.assert_that_pv_is_number(UNDERLYING_GAP_SP.format(jaw, direction), expected_gaps[jaw - 1], 0.1)
 
+    def _test_WHEN_sizes_at_moderator_and_sample_changed_THEN_centres_of_all_jaws_unchanged(self, direction):
```
DominicOram: This would be a better test if each jaw had a different centre at the start. This test would still pass if the manager did something stupid like set all centres to that of the first jaw.

Tom-Willemsen: Done, thanks

PR #228, opened 2019-09-02
Ticket4011 edwards tic

Alistair-McGann-Tessella:
Adds tests for the edwards turbo instrument controller

Ticket:
https://github.com/ISISComputingGroup/IBEX/issues/4011

### DIFF
--- a/tests/edwardstic.py
+++ b/tests/edwardstic.py
@@ -232,7 +232,7 @@ class EdwardsTICTests(unittest.TestCase):
     def setUp(self):
         self._lewis, self._ioc = get_running_lewis_and_ioc("edwardstic", DEVICE_PREFIX)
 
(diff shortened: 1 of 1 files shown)


tests/edwardstic.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+class Gauge2Tests(GaugeTestBase, unittest.TestCase):
+    def get_gauge_number(self):
+        return 2
+
+
+class Gauge3Tests(GaugeTestBase, unittest.TestCase):
+    def get_gauge_number(self):
+        return 3
```
Tom-Willemsen: Usually we would just do this with `@parameterized.expand`, maybe in combination with `itertools.product` if we also want to parameterize some other parameter at the same time. This is fine here though and does not need changing - just a note for future.

PR #233, opened 2019-09-16
Tests for 361 added to check that they only work with channel 1

JamesKingWork:
https://github.com/ISISComputingGroup/IBEX/issues/4680


tests/tpg361.py
```diff
@@ -0,0 +1,122 @@
...
+TEST_MODES = [TestModes.RECSIM, TestModes.DEVSIM]
+
+
+class Tpg361Tests(TpgBase, unittest.TestCase):
+    def get_prefix(self):
+        return DEVICE_PREFIX
+
+    @parameterized.expand(("Pressure_{}".format(i), i) for i in range(1, 3))
+    def test_GIVEN_pressure_set_WHEN_read_THEN_pressure_is_as_expected(self, _, channel):
+        if channel == 1:
```
Tom-Willemsen: We're parametrizing the whole test for 3 channels but then only actually testing one of them. I think we can remove the parameterized decorator and the `if`-case in this test.

Comment repeats for all similar cases

PR #236, opened 2019-09-20
Remove submodule checkout on build

John-Holt-Tessella:
Remove the need for ibex utils to be a submodule. Instead of build check out the lastest version.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,7 +24,7 @@ pipeline {
     stage("Install latest IBEX") {
       steps {
         bat """
-            if EXIST "ibexutils" rmdir /s /q ibex_utils
+            if EXIST "ibex_utils" rmdir /s /q ibex_utils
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -24,6 +24,8 @@ pipeline {
     stage("Install latest IBEX") {
       steps {
         bat """
+            if EXIST "ibexutils" rmdir /s /q ibex_utils
```
Tom-Willemsen:
```suggestion
            if EXIST "ibex_utils" rmdir /s /q ibex_utils
```

Jenkinsfile7
```diff
@@ -24,6 +24,8 @@ pipeline {
     stage("Install latest IBEX") {
       steps {
         bat """
+            if EXIST "ibexutils" rmdir /s /q ibex_utils
```
Tom-Willemsen:
```suggestion
            if EXIST "ibex_utils" rmdir /s /q ibex_utils
```

PR #237, opened 2019-09-22
Fix submodule issue 

FreddieAkeroyd:
* remove .gitmodules
* Remove ibex_utils after install, if it is not needed for tests it should be removed as developers will not have it present when tests are run manually


FreddieAkeroyd:
See https://github.com/ISISComputingGroup/EPICS-IOC_Test_Framework/pull/236

PR #243, opened 2019-10-01
IOCTestFramework: Add BeckhoffEmulatorLauncher check for missing AutomationTools binary

aaron-long:
Added a check to ensure the user has built the `AutomationTools` binary. Gives feedback if they have failed to do this.


aaron-long:
`IOError` raised if binary is not built.

PR #292, opened 2020-02-06
added test for voltage limits

Tom-Willemsen:
https://github.com/ISISComputingGroup/IBEX/issues/5104


Tom-Willemsen:
@ChrisColeExpControl we now have a debug build of system_tests_iocs on the build server which should point out a problem if it is related to debug builds. Given that this issue is only reproducible on some machines I'm not sure it's going to be possible to debug locally. 

If that is the only thing blocking a merge, can I suggest merging and then keeping an eye on the system_tests_ioc_debug build to see whether the problem is wider or isolated to one or two dev machines?

Alternatively, we can look together and try to step through it in a debugger on your machine to try to pin down where your machine differs from mine and the build server.

PR #306, opened 2020-03-20
Ticket 4287: update from tests

DominicOram:
See https://github.com/ISISComputingGroup/IBEX/issues/4287

### DIFF
--- a/tests/rotsc.py
+++ b/tests/rotsc.py
@@ -5,8 +5,8 @@ from parameterized import parameterized
 from utils.channel_access import ChannelAccess
 from utils.ioc_launcher import get_default_ioc_dir
 from utils.test_modes import TestModes
-from utils.testing import IOCRegister, skip_if_recsim, get_running_lewis_and_ioc, assert_log_messages, \
-    parameterized_list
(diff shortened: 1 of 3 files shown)


tests/rotsc.py
```diff
@@ -63,3 +85,60 @@ def test_GIVEN_current_position_WHEN_position_set_to_current_position_THEN_setpo
         # WHEN
```
ChrisColeExpControl, line 85: This test is now failing consistently on my machine due to the new log messages being generated when switching states, as assert_pv_processed is looking for a log message which gets split up by them.
Failure is happening on line 81 but github won't let me comment there.

DominicOram: Annoyingly I can't reproduce this on mine. Can I see your `log_ROTSC_01_devsim_ioc.log`?

DominicOram: Ok, think I've updated the process checking to be more robust

DominicOram: Can you try it now along with https://github.com/ISISComputingGroup/genie_python/pull/245?

PR #312, opened 2020-03-26
Add new argument to just run recsim or devsim

FreddieAkeroyd:
See ISISComputingGroup/IBEX#5349


FreddieAkeroyd:
To test:
```
-tm DEVSIM should only run devsim tests
-tm RECSIM should only run recsim tests
not specified should run both as before
```

PR #319, opened 2020-04-09
Ticket5259 switch rates mid ramp

ChrisColeExpControl:
### Description off work
Added tests for ramp rate changing based on the table found in cryosms/ramps/. A variety of start and end points are tested. Additionally, previous tests relied on placeholder code in the support files which set targets to an arbitrarily higher value whenever a ramp was requested. This placeholder code has been replaced  with the actual code for ramp requests, so we now need to set targets in the old tests.
Furthermore, the setup phase of testing has been vastly improved to ensure that both the device and the driver are returned to an initial state after each test, no matter the result of the test.

### To test
[Ticket 5259](https://github.com/ISISComputingGroup/IBEX/issues/5259)

### Acceptance criteria
 - New tests adequately test the acceptance criteria in the above ticket
 - Code is of an acceptable quality
 - Old tests modified as necessary to ensure their continued functionality


John-Holt-Tessella (approved):
All good now

PR #325, opened 2020-04-30
Ticket5365

Tom-Willemsen:
https://github.com/ISISComputingGroup/IBEX/issues/5365


Tom-Willemsen:
I don't think there's currently a way to detect reconnection from the lewis side.

I think the test `test_WHEN_emulator_sends_corrupt_packets_THEN_records_go_into_alarm` is trying to do the workflow you describe as it asserts that the `RESET_CONNECTION` pv was processed when corrupt messages are received and then sets `corrupted_messages` back to `False` via the backdoor? Or have I misunderstood what you meant? Happy to have a chat if that's easier...

John-Holt-Tessella:
Test looks good and you can't easily test a reconnect

PR #328, opened 2020-05-09
Create specific named function for failed test

FreddieAkeroyd:
Create name of error reporting function dynamically rather than use default runTest name, this is to improve error report summary page that just shows test name (you have to click for details)

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -7,6 +7,7 @@ import os
 import sys
 import traceback
 import unittest
+import types
 
 import six
 import xmlrunner
@@ -128,7 +129,7 @@ def load_and_run_tests(test_names, failfast, ask_before_running_tests, tests_mod
             clean_environment()
             device_launchers = make_device_launchers_from_module(module.file, mode)
             test_results.append(
-                run_tests(arguments.prefix, module.tests, device_collection_launcher(device_launchers),
+                run_tests(arguments.prefix, module.name, module.tests, device_collection_launcher(device_launchers),
                           failfast, ask_before_running_tests))
 
     return all(test_result is True for test_result in test_results)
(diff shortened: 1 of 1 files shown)


run_tests.py
```diff
@@ -155,30 +156,36 @@ def prompt_user_to_run_tests(test_names):
...
+        msg:         Error message explaining test failure.
 
     Returns:
         None
     """
-    def __init__(self, msg):
-        super(ReportFailLoadTestsuiteTestCase,self).__init__()
+    def __init__(self, func_name, msg):
+        setattr(self, func_name, types.MethodType(runTestProto, self))
+        super(ReportFailLoadTestsuiteTestCase,self).__init__(func_name)
```
John-Holt-Tessella, line 179: This is setting the method name to run to be the same as function that doesn't seem right. Also unit test says:

```    If it is necessary to override the __init__ method, the base class
    __init__ method must always be called. It is important that subclasses
    should not change the signature of their __init__ method, since instances
    of the classes are instantiated automatically by parts of the framework
    in order to be run.
```

I think we should not create out own code. Although maybe you have already had this argument.

run_tests.py
```diff
@@ -155,30 +156,36 @@ def prompt_user_to_run_tests(test_names):
         elif answer.upper()[0] == "Y":
             return
 
+def runTestProto(self):
```
John-Holt-Tessella, line 159: move this in the class

run_tests.py
```diff
@@ -155,30 +156,36 @@ def prompt_user_to_run_tests(test_names):
...
+        func_name:   Name of function to create for the test.
+        msg:         Error message explaining test failure.
 
     Returns:
         None
     """
-    def __init__(self, msg):
-        super(ReportFailLoadTestsuiteTestCase,self).__init__()
+    def __init__(self, func_name, msg):
+        setattr(self, func_name, types.MethodType(runTestProto, self))
```
John-Holt-Tessella, line 178: change func_name to failing_module_name then
func_name = "test_start_{}_test_environment_and_ioc"
setattr(self, func_name, self.runTestProto)

run_tests.py
John-Holt-Tessella: Acrtually probably ok but leave a coment saying this is only use din our ioc testing framework

run_tests.py
John-Holt-Tessella: rename to fail_with_message

run_tests.py
John-Holt-Tessella: You might be able to get away with not using init(funcname

run_tests.py
John-Holt-Tessella: Might not have to do this at all if the function you add is called test_xxx

run_tests.py
FreddieAkeroyd: I originally had it as a class mehtod, but it didn't like it as you have to pass the name of the function to super init and it seemed to think it didn't exist at that point. But maybe you have to do something cleverer with the syntax?

FreddieAkeroyd: I mean pass it to setattr so that you can then pass the name to super init

run_tests.py
FreddieAkeroyd: The function name (failed module name) need to pass to super init, so I passed it by subclass init - is there any other way other than a class variable?

run_tests.py
FreddieAkeroyd: > You might be able to get away with not using init(funcname
Sorry, I don't follow - or is the suggestion to create a new class dynamically instead so the  function nane doesn't need to change?

run_tests.py
FreddieAkeroyd: Well, it now seems to work as a class method - maybe I had setattr wrong first time or something

PR #336, opened 2020-06-08
Fix ioc tests

Tom-Willemsen:
- Reverts https://github.com/ISISComputingGroup/EPICS-IOC_Test_Framework/commit/2b0c1542cce6c1aacc5b6211341cd6539f327299 because the build server fails unless we use an explicit path to python in the startup command
- `g.get_instrument()` -> `g.adv.get_instrument()`

### DIFF
--- a/tests/danfysik_local_calib.py
+++ b/tests/danfysik_local_calib.py
@@ -30,7 +30,7 @@ class DanfysikLocalCalibTests(DanfysikBase, unittest.TestCase):
     """
     def test_GIVEN_local_calib_macro_set_to_no_THEN_calib_base_dir_is_common_dir(self):
         g.set_instrument(None)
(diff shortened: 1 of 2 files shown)


DominicOram:
I originally did https://github.com/ISISComputingGroup/EPICS-IOC_Test_Framework/commit/2b0c1542cce6c1aacc5b6211341cd6539f327299 because you can't run `python.exe lewis.exe` if you install lewis via the `tar`. If we can't run the `lewis.exe` on it's own shall we do `python -m lewis`?

Tom-Willemsen:
Ok yeah I think `python -m lewis` would work

DominicOram:
I don't think we ever do. I was doing it to when I was testing the latest `lewis` release

PR #339, opened 2020-06-25
Ticket 5412: Add Danfysik test that restarts IOC

DominicOram:
See https://github.com/ISISComputingGroup/IBEX/issues/5412

To test:
* Run some of the following kinds of tests and confirm they still pass:
    * `ProcservLauncher` tests (`simple` or `lsicorr` use this)
    * `PythonIOCLauncher` tests (`refl`)
    * `IOCLauncher` tests (any other tests)
* Switch some `IOCLauncher` tests to be `ProcservLauncher` tests and confirm they still pass
* Run all new danfysik tests and confirm they pass


DominicOram:
> The harmonisation makes the code nicer. The danfysik tests do not pass, it seems that the data on the emulated device is reinitialised when the IOC is restarted so the value of power is not what is expected.

The tests all pass for me. I don't see it calling reinitialise during the test, only when the test first starts. If you change the test to:
```python
# @parameterized.expand([
    #     ("power_on_and_current_at_10",  True, 10),
    #     ("power_off_and_current_at_50", False, 50),
    # ])
    @skip_if_recsim("In rec sim this test fails as there is nothing holding the device state")
    def test_WHEN_IOC_is_restarted_THEN_current_and_powered_are_not_changed(self):
        power_state = True
        current = 10
```
then run it on it's own like `python run_tests.py -t danfysik8000.Danfysik8000Tests.test_WHEN_IOC_is_restarted_THEN_current_and_powered_are_not_changed -tm DEVSIM` does it still fail? Do you see it reinitialise the emulator more than the once at the beginning?

PR #346, opened 2020-07-17
increase timeout for jasco

Tom-Willemsen:
Various tests are failing regularly on the build server (but not 100% of the time). See if increasing the general timeout helps.


Alexandru-Dascalu:
Looks good!

PR #347, opened 2020-07-17
Add tests to check error status set correctly

JamesKingWork:
https://github.com/ISISComputingGroup/IBEX/issues/4578

### DIFF
--- a/tests/sans2d_vacuum.py
+++ b/tests/sans2d_vacuum.py
@@ -175,7 +175,7 @@ class Sans2dVacuumTankTest(unittest.TestCase):
         self._ioc = IOCRegister.get_running("FINS_01")
         self.ca = ChannelAccess(device_prefix=ioc_prefix)
 
-    @parameterized.expand(parameterized_list([-5, 0, 3, 5, 7, 9]))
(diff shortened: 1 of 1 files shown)


tests/sans2d_vacuum.py
```diff
@@ -163,3 +163,25 @@ def test_WHEN_end_run_in_manual_shutter_mode_THEN_shutter_opened(self):
...
+class Sans2dVacuumTankTest(unittest.TestCase):
+    """
+    Tests for the SANS2D vacuum tank, based on a FINS PLC.
+    """
+
+    def setUp(self):
+        self._ioc = IOCRegister.get_running("FINS_01")
+        self.ca = ChannelAccess(device_prefix=ioc_prefix)
+
+    @parameterized.expand([(-5,), (0,), (3,), (5,), (6,), (7,), (9,), (10,)])
```
Alexandru-Dascalu, line 177: 1) I do not think testing for 6 and 10 is worth it, as 6 is between 5 and 7 so that interval is covered, and if 9 passes then there is no reason 10 should not, they are in the same segment of invalid values.
2) I think you should use parameterized_list to make the code consistent with other tests and so you do not need to have that ugly list.

PR #361, opened 2020-08-18
Stabilize jasco tests

Tom-Willemsen:
https://github.com/ISISComputingGroup/IBEX/issues/5630

### DIFF
--- a/tests/jsco4180.py
+++ b/tests/jsco4180.py
@@ -22,7 +22,6 @@ IOCS = [
 
 TEST_MODES = [TestModes.DEVSIM]
 
-
 ERROR_STATE_HARDWARE_FAULT = 4
 ERROR_STATE_NO_ERROR = 2
 
@@ -34,6 +33,7 @@ class Jsco4180Tests(unittest.TestCase):
     """
     Tests for the Jsco4180 IOC.
     """
+
     def setUp(self):
         self._lewis, self._ioc = get_running_lewis_and_ioc(DEVICE_NAME, DEVICE_PREFIX)
         self.ca = ChannelAccess(device_prefix=DEVICE_PREFIX, default_timeout=30)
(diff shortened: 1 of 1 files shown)


tests/jsco4180.py
```diff
@@ -79,21 +79,33 @@ def test_GIVEN_wrong_component_on_device_WHEN_send_new_sp_THEN_state_machine_abo
...
-        # however if we change setpoint, the loop should start again
-        self.ca.set_pv_value("COMP:A:SP", value-10)
-        self.ca.set_pv_value("COMP:B:SP", value+10)
-        sleep(delay / 2.0 + 1.0) # reset should not have happened
-        self.ca.assert_that_pv_is("COMP:A", value-5)
-        sleep(delay / 2.0) # reset should now have happened
-        self.ca.assert_that_pv_is("COMP:A", value-10, timeout=15)
+        initial_delay = self.ca.get_pv_value("ERROR:DELAY")  # delay before state machine reset
+        delay = 30  # Increase delay to avoid race conditions
+        self.ca.set_pv_value("ERROR:DELAY", delay)
```
Alistair-McGann-Tessella, line 84: I'm not sure `ERROR:DELAY` is ever actually used. This might be adding an arbitrary sleep?

tests/jsco4180.py
```diff
@@ -79,21 +79,33 @@ def test_GIVEN_wrong_component_on_device_WHEN_send_new_sp_THEN_state_machine_abo
...
+        try:
+            with self.ca.assert_pv_not_processed("RESET:SP", timeout=1):
+                self._lewis.backdoor_set_on_device("component_A", value - 5)
+                self.ca.assert_that_pv_is("COMP:A", value - 5, timeout=5)
+                sleep(delay / 2.0)
+
+                # however if we change setpoint, the loop should start again
+                self._lewis.backdoor_set_on_device("component_A", value - 5)
```
Alistair-McGann-Tessella, line 92: Does this line do anything? `component_A` should already equal `value - 5` from above?

tests/jsco4180.py
Tom-Willemsen: It is used internally to the jasco ioc by it's state machine. We want it to delay for longer otherwise we get races where the state machine in the ioc is faster than the framework

tests/jsco4180.py
Tom-Willemsen: No I think you're right it can be removed if the test still passes with it removed

Tom-Willemsen:
I ran the test 20 times locally and was unable to reproduce the error. I've put on an unstable test decorator with a max retry count of 2 just in case it does crop up occasionally though.

John-Holt-Tessella (approved):

PR #366, opened 2020-09-02
Modify to handle both protocols

Tom-Willemsen:
https://github.com/ISISComputingGroup/IBEX/issues/5545


JamesKingWork (approved):
Matches old code

PR #367, opened 2020-09-08
Ticket 5356 - Tests check for non existant PV's

aaron-long:
Updated various tests to check valid PV's for their existence when starting the driver:

- SM300 does not have `DISABLE`: changed to `AXIS1` (known record)
- READASCII not have `DISABLE`: changed to `DIRBASE` (known record)
- motionSetpoints does not have `DISABLE`. pv_for_existence uses NAME but there is no motion_setpoints: Added a custom prefix `MSP`.
- GAMRY does not have `DISABLE`: Changed to `CHARGE:STAT` (known record)


DominicOram (approved):
Great, thanks!

PR #369, opened 2020-09-24
Add test for DAQMX reconnection in monster mode

Tom-Willemsen:
https://github.com/ISISComputingGroup/IBEX/issues/5752

Add test for DAQMX reconnection in monster mode


DominicOram (changes requested):
As discussed, I think it would be better if `DAQmxMonsterTests` and `DAQmxTests` use a common base class, I will implement this now.

DominicOram (approved):
Tests work well, thanks.

PR #372, opened 2020-10-07
Add tests for taking measurements and writing data to PV.

JamesKingWork:
https://github.com/ISISComputingGroup/IBEX/issues/5633


DominicOram (changes requested):
I'm getting a failing test:
image
which I believe is because the correlator is expecting FILEPATH to be folder rather than a file? However, when I change it to be a folder it gives:
image

When I comment out this test others work fine. 

Additionally when I'm getting these errors the IOC test framework does not correctly shutdown the IOC so I have to do it manually before running the framework again. This may be an issue with procserv IOCs in the test framework in general so outside the scope of this ticket. Feel free to make a new ticket.

DominicOram (approved):
Good tests, thanks

PR #376, opened 2020-10-14
Ticket3185 remote mode reliability

JamesKingWork:
### Description of work

- Split tests out into kepco versions with SYST:REM command available and without (by setting the firmware at the start of each test which is the condition the device emulator checks)
- Test firmware and IDN are retrieved correctly with multiple versions
- Test when kepco IOC restarted and the firmware version and macro satisfy the conditions to do so then a reset is sent and parameters are also resent
- Test when kepco IOC restarted and the firmware version and macros satisfy the conditions to do so then a remote command is sent
- Add ability to start an IOC with a given set of macros and the ability to start again with the original macros resetting them to the expected 
- Add ability to start an IOC and wait for the given startup text
- Add assertion that a dictionary with keys as PVs and values as expected values have the given expected values

### To test

https://github.com/ISISComputingGroup/IBEX/issues/3185

### Acceptance criteria

*List the acceptance criteria for the PR*

---

#### Code Review

- [ ] A copy of the manual has been placed on the shared drive
- [ ] Pertitent information has been stored in the [wiki](https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/Specific-Device-IOC)
- [ ] Does the IOC conform to IBEX standards?
    - [PV naming](https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/PV-Naming).
    - [Disable records](https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/Disable-records)
    - [Record simulation](https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/Record-Simulation)
    - [Finishing touches](https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/IOC-Finishing-Touches)
- [ ] If an OPI has been modified, does it conform to the [style guidelines](https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/OPI-Creation)?
- [ ] Do the IOC and support module conform to the new [build guidelines](https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/Reducing-Build-Dependencies)
- [ ] Have the changes been recorded appropriately in the [release notes](https://github.com/ISISComputingGroup/IBEX/wiki/ReleaseNotes_Dev)?
- [ ] Is the device's flow control setting correct? [For most devices flow control should be OFF](https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/Flow-control).

### Functional Tests

- IOC responds correctly in:
    - [ ] Devsim mode
    - [ ] Recsim mode
    - [ ] Real device, if available
- [ ] Supplementary IOCs (`..._0n` where `n>1`) run correctly
- [ ] Log files do not report undefined macros (serach for `macLib: macro` to find instances of `macLib: macro [macro name] is undefined...`

### Final steps

- [ ] Update the IOC submodule in the main EPICS repo. See [Git workflow](https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/Git-workflow) page for details.
- [ ] Reviewer has moved the [release notes](https://github.com/ISISComputingGroup/IBEX/wiki/ReleaseNotes_Dev) entry for this ticket in the "Changes merged into master" section

### DIFF
--- a/tests/kepco_no_rem.py
+++ b/tests/kepco_no_rem.py
@@ -51,7 +51,7 @@ class KepcoNoRemTests(KepcoTests, unittest.TestCase):
             try:
                 self._lewis.assert_that_emulator_value_is(var, set_func(val), cast=int)
             except AssertionError as e:
-                error_message_calls += "\n{}".format(e.message)
+                error_message_calls += "\n{}\n{}".format(var, e.message)
         if error_message_calls != "":
             raise AssertionError("Failed to call sets:{}".format(error_message_calls))
 
@@ -92,7 +92,9 @@ class KepcoNoRemTests(KepcoTests, unittest.TestCase):
         )
 
         # Restart the ioc, initiating a reset and sending of values after 100 microseconds
-        self._ioc.start_with_macros(macros)
+        self._ioc.start_with_macros(macros, wait=True)
+        self.ca.assert_that_pv_exists("VOLTAGE", timeout=60)
+
 
         # Assert reset occurred
         self._lewis.assert_that_emulator_value_is("reset_count", 1, cast=int)
--- a/tests/kepco_no_rem.py
+++ b/tests/kepco_no_rem.py
@@ -107,4 +107,7 @@ class KepcoNoRemTests(KepcoTests, unittest.TestCase):
 
         # Assert autosave has correctly set pvs
         self.ca.assert_dict_of_pvs_have_given_values(pv_values)
-        self._ioc.start_with_original_macros()
+
+        # Reset the macros for subsequent tests
+        self._ioc.start_with_original_macros(wait=True)
+        self.ca.assert_that_pv_exists("VOLTAGE", timeout=60)
--- a/tests/kepco_no_rem.py
+++ b/tests/kepco_no_rem.py
@@ -18,7 +18,6 @@ IOCS = [
         "directory": get_default_ioc_dir("KEPCO"),
         "emulator": emulator_name,
         "ioc_launcher_class": ProcServLauncher,
-        "lewis_protocol": "no_rem",
     },
 ]
 
@@ -58,9 +57,9 @@ class KepcoNoRemTests(KepcoTests, unittest.TestCase):
 
     @parameterized.expand(parameterized_list([
         (IDN_NO_REM[0], IDN_NO_REM[1], {}),
-        (IDN_NO_REM[0], IDN_NO_REM[1], {"ON_START": 0}),
-        (IDN_NO_REM[0], IDN_NO_REM[1], {"ON_START": 1}),
-        (IDN_REM[0], IDN_REM[1], {"ON_START": 1}),
+        (IDN_NO_REM[0], IDN_NO_REM[1], {"RESET_ON_START": 1}),
+        (IDN_NO_REM[0], IDN_NO_REM[1], {"RESET_ON_START": 2}),
+        (IDN_REM[0], IDN_REM[1], {"RESET_ON_START": 2}),
     ]))
     @skip_if_recsim("Lewis not available in recsim")
     def test_GIVEN_kepco_started_THEN_reset_and_params_resent(self, _, idn_no_firmware, firmware, macros):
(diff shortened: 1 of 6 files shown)


tests/kepco_rem.py
```diff
@@ -0,0 +1,89 @@
...
+
+        self._ioc.start_with_macros(macros, wait=True)
+        self.ca.assert_that_pv_exists("VOLTAGE", timeout=60)
+
+        self._lewis.assert_that_emulator_value_is("remote_comms_enabled", True, cast=strtobool)
+        self._lewis.assert_that_emulator_value_is("reset_count", 0, cast=int)
+
+        # Reset the macros for subsequent tests
+        self._ioc.start_with_original_macros(wait=True)
+        self.ca.assert_that_pv_exists("VOLTAGE", timeout=60)
```
Tom-Willemsen, line 71: I am worried that if this test fails these lines won't get run and it will then corrupt the state for the next test(s). I think it would be better to put these in setup so that they always run before each test? Or alternatively we need to do something like:
```
def test():
    try:
        # the actual test logic
    finally:
        self._ioc.start_with_original_macros(wait=True)
        self.ca.assert_that_pv_exists("VOLTAGE", timeout=60)
```
could also implement a decorator/context manager to do this for us to avoid duplication.

comment repeats as necessary for similar pattern

tests/kepco_no_rem.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Set counts to zero and remote mode to false
+        # We want lewis_vars and lewis_vals separate as they are checked later
+        lewis_vars = ["voltage_set_count", "current_set_count", "output_mode_set_count", "output_status_set_count"]
+        lewis_vals = [0, 0, 0, 0]
+        self.lewis_set_and_assert_list(
+            zip(lewis_vars + ["reset_count", "remote_comms_enabled"], lewis_vals + [0, False])
+        )
+
+        # Restart the ioc, initiating a reset and sending of values after 100 microseconds
```
Tom-Willemsen, line 94: I don't understand what is meant by this comment - from the code it looks like it resends using PINI? Where does the 100 microseconds come into this?

Tom-Willemsen (changes requested):
Good coverage for the case where the IOC is restarted. However we also have the option to process the `RSTRSND` PV from the GUI (without restarting the ioc) and this appears to be untested as I can't find any reference to `RSTRSND` in this PR. Can we add a test for this case please?

tests/kepco_no_rem.py
JamesKingWork: There's a wait 100 in the proto after the reset, then the values are resent

tests/kepco_rem.py
```diff
@@ -0,0 +1,77 @@
...
+        self._set_IDN(idn_no_firmware, firmware)
+        self._lewis.backdoor_set_and_assert_set("reset_count", 0)
+        self._lewis.backdoor_set_and_assert_set("remote_comms_enabled", False)
+
+        with self._ioc.start_with_macros(macros, "VOLTAGE"):
+            self._lewis.assert_that_emulator_value_is("remote_comms_enabled", True, cast=strtobool)
+            self._lewis.assert_that_emulator_value_is("reset_count", 0, cast=int)
+
+    @skip_if_recsim("Lewis not available in recsim")
+    def test_GIVEN_kepco_firmware_does_not_support_SYSTREM_WHEN_on_start_is_2_THEN_no_remote_mode_AND_no_reset(self):
```
Tom-Willemsen, line 68: I can't get this test to pass:

```
Running tests...
----------------------------------------------------------------------
..................F............
======================================================================
FAIL [20.202s]: test_GIVEN_kepco_firmware_does_not_support_SYSTREM_WHEN_on_start_is_2_THEN_no_remote_mode_AND_no_reset (tests.kepco_rem.KepcoRemTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\Instrument\Apps\EPICS\support\IocTestFramework\master\utils\testing.py", line 147, in wrapper
    return func(*args, **kwargs)
  File "c:\Instrument\Apps\EPICS\support\IocTestFramework\master\tests\kepco_rem.py", line 75, in test_GIVEN_kepco_firmware_does_not_support_SYSTREM_WHEN_on_start_is_2_THEN_no_remote_mode_AND_no_reset
    self._lewis.assert_that_emulator_value_is("remote_comms_enabled", False, cast=strtobool)
  File "c:\Instrument\Apps\EPICS\support\IocTestFramework\master\utils\emulator_launcher.py", line 217, in assert_that_emulator_value_is
    emulator_property, lambda val: cast(val) == expected_value, timeout=timeout, msg=message)
  File "c:\Instrument\Apps\EPICS\support\IocTestFramework\master\utils\emulator_launcher.py", line 255, in assert_that_emulator_value_causes_func_to_return_true
    raise AssertionError(err)
AssertionError: Expected PV to have value 'False' (type: 'bool').
Final emulator property value was 'True' (type: 'str')
```

Tom-Willemsen (changes requested):

Tom-Willemsen (approved):

PR #383, opened 2020-10-28
Ticket 5662: Added tests for Kepco magnet calibration

DominicOram:
See https://github.com/ISISComputingGroup/IBEX/issues/5662

* Moved generic setting of calibration files out of eurotherm and into utils
* Added tests for this in the kepco

### DIFF
--- a/tests/kepco.py
+++ b/tests/kepco.py
@@ -7,6 +7,7 @@ from utils.ioc_launcher import get_default_ioc_dir
 from utils.test_modes import TestModes
 from utils.testing import get_running_lewis_and_ioc, skip_if_recsim, parameterized_list
 from distutils.util import strtobool
+from utils.calibration_utils import use_calibration_file, reset_calibration_file
 
 DEVICE_PREFIX = "KEPCO_01"
 
@@ -51,6 +52,7 @@ class KepcoTests(unittest.TestCase):
         self.ca = ChannelAccess(default_timeout=30, device_prefix=DEVICE_PREFIX)
         self._lewis.backdoor_run_function_on_device("reset")
         self.ca.assert_that_pv_exists("VOLTAGE", timeout=30)
+        reset_calibration_file(self.ca)
 
     def _write_voltage(self, expected_voltage):
         self._lewis.backdoor_set_on_device("voltage", expected_voltage)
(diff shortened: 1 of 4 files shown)


tests/kepco.py
```diff
@@ -139,3 +141,44 @@ def test_GIVEN_psu_in_local_mode_WHEN_setpoint_is_sent_THEN_power_supply_put_int
...
+    def test_GIVEN_calibration_of_none_WHEN_current_read_THEN_field_equal(self):
+        test_current = 100
+        expected_field = 100
+
+        with use_calibration_file(self.ca, "None.txt"):
+            self._write_current(test_current)
+            self.ca.assert_that_pv_is("FIELD", expected_field)
+
+    @skip_if_recsim("Calibration lookup does not work in recsim")
+    def test_GIVEN_calibration_of_none_WHEN_current_read_THEN_field_equal(self):
```
rerpha, line 176: Perhaps a little bit over the top, but could this be parametrized with the file name and a tuple of the expected current and fields? or list of tuples, then you can test multiple values

utils/calibration_utils.py
```diff
@@ -0,0 +1,32 @@
...
+def set_calibration_file(channel_access, filename):
+    """
+    Sets a calibration file. Retries if it didn't set properly first time.
+    """
+    max_retries = 10
+
+    for _ in range(max_retries):
+        channel_access.set_pv_value("CAL:SEL", filename)
+        channel_access.assert_that_pv_alarm_is("CAL:SEL", channel_access.Alarms.NONE)
+        time.sleep(3)
```
rerpha, line 14: can we process the pv here instead of manually sleeping?

utils/calibration_utils.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def set_calibration_file(channel_access, filename):
+    """
+    Sets a calibration file. Retries if it didn't set properly first time.
+    """
+    max_retries = 10
+
+    for _ in range(max_retries):
+        channel_access.set_pv_value("CAL:SEL", filename)
```
rerpha, line 12: I think "CAL:SEL" should be moved out into a constant here in case we ever change the pv name

rerpha (commented):
tests look good, some minor comments. We should make sure the test framework copies in the `None.txt` file into `config/calib` for testing

utils/calibration_utils.py
DominicOram: Good suggestion but in this case I'm not 100% sure where the slow down is coming from so I think we'll err on the side of stability over speed.

rerpha (approved):

PR #390, opened 2020-11-27
Set correct PV into alarm

JamesKingWork:
System tests were failing with no alarm. This was because the incorrect pv was being set into alarm to simulate an invalid power supply. CURRENT:SP:RBV takes alarm from _CURRENT:SP:RBV so set that into alarm.


DominicOram (approved):
Oops, that was probably my fault with the KEPCO stuff.

PR #401, opened 2020-12-22
Remove EPICS dir at end

DominicOram:
Having separate EPICS directories is clogging up RENO, this should free up some space.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -119,7 +119,7 @@ pipeline {
 	stage("Clean Up") {
       steps {
 	    bat """
-		  rmdir "C:\\Instrument\\Apps\\EPICS-%MYJOB%"
+		  rmdir /s /q "C:\\Instrument\\Apps\\EPICS-%MYJOB%"
 		  exit /b %ERRCODE%
 		"""
 	  }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -115,6 +115,15 @@ pipeline {
          }
       }
     }
+	
+	stage("Clean Up") {
+      steps {
+	    bat """
+		  rmdir "C:\\Instrument\\Apps\\EPICS-%MYJOB%"
+		  exit /b %ERRCODE%
```
FreddieAkeroyd, line 123: I think you need /s /q as this is a real directory, also probably a good idea to remove the instrument/apps/epics junction too and run stop_ibex_server first in case any processes using files. As this is a sort of repeat of what is done in the install section, maybe create a function instead for both to call?

JamesKingWork: Saw this, as well, added

JamesKingWork (approved):

Jenkinsfile
DominicOram: Thanks both, Freddie is right about running stop and removing the junction though. I'll put in another PR.

FreddieAkeroyd: Also may be worth moving to a post section instead - if tests fail it may not run the next stage as written

PR #405, opened 2021-01-04
Telnetlib calls must use bytes

FreddieAkeroyd:
Make sure data to/from telnetlib is encoded/treated as bytes


John-Holt-Tessella (approved):
Hope this is ok change

PR #406, opened 2021-01-08
Fix tests

DominicOram:
Fixes the danfysik IOC system tests. Note that not converting the Lewis return value to a string may break other tests but we should use the type properly now that we can, I suggest we merge and see what fails on the build server.

To test:
* Run the danfysik tests and confirm they pass

### DIFF
--- a/utils/emulator_launcher.py
+++ b/utils/emulator_launcher.py
@@ -465,7 +465,7 @@ class LewisLauncher(EmulatorLauncher):
         :return: lines from the command output
         """
         try:
-            return str(call_method(self.remote, lewis_command[0], lewis_command[1], lewis_command[2:]))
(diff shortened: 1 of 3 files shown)


utils/emulator_launcher.py
```diff
@@ -493,7 +493,7 @@ def backdoor_get_from_device(self, variable_name, *_, **__):
         """
         Return the string of a value on a device from lewis.
         :param variable_name: name of the variable
-        :return: the variables value, as a string
+        :return: the variables value
         """
         # backdoor_command returns a list of bytes and join takes str so convert them here
```
FreddieAkeroyd, line 498: I was a little confused by this comment - does `str` here refer to the `str(variable_name)` in this command, or does it refer to the `str` that has been removed from `backdoor_command` above?

PR #415, opened 2021-02-11
Fix knr1050 tests and speed them up

DominicOram:
Fixes knr1050 tests and implements https://github.com/ISISComputingGroup/IBEX/issues/6225.

To test:
* Confirm knr1050 tests pass, are stable and faster than on master

### DIFF
--- a/tests/knr1050.py
+++ b/tests/knr1050.py
@@ -29,7 +29,7 @@ class Knr1050Tests(unittest.TestCase):
     """
     def setUp(self):
         self._lewis, self._ioc = get_running_lewis_and_ioc(device_name, DEVICE_PREFIX)
-        self.ca = ChannelAccess(device_prefix=DEVICE_PREFIX)
+        self.ca = ChannelAccess(device_prefix=DEVICE_PREFIX, default_wait_time=0.0)
         self._lewis.backdoor_run_function_on_device("reset")
         # Set the device in remote mode ready to receive instructions
         self.ca.set_pv_value("MODE:SP", "REMOTE")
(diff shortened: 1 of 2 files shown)


tests/knr1050.py
```diff
@@ -86,7 +86,7 @@ def test_GIVEN_set_concentration_via_ioc_WHEN_ramp_command_sent_via_ioc_THEN_cor
         self.ca.set_pv_value("COMP:B:SP", expected_concentrations[1])
         self.ca.set_pv_value("COMP:C:SP", expected_concentrations[2])
         self.ca.set_pv_value("COMP:D:SP", expected_concentrations[3])
-        self.ca.set_pv_value("START:SP", 1)
+        self.ca.set_pv_value("START:SP", 1, sleep_after_set=1.0)
```
FreddieAkeroyd, line 90: Would it be clearer to use `sleep(1.0)` with a comment that a `sleep` was needed for the emulator to process the data - I feel adding sleeps inside "set pv" is hiding things and making it less clear why they are being done. Also having a default sleep mechanism inside a `set_pv_value` only used during testing not only slows down the tests, but also potentially hides race conditions. Previously the four concentration commands here would have had a 1 second sleep automatically added in tests, but not in a user script, so if there had been a race condition in db logic (which there isn't) the tests would have missed it.

DominicOram: Agreed, we could add the `sleep(1.0)` here and work towards removing the sleep from `set_pv_value` eventually but I want to do it yet 'cos it will just break all the tests again

FreddieAkeroyd: yes, so the plan would be to move towards having tests creating a CaChannel with "default_wait_time=0.0" (it would still be 1.0 if not specified) and have any needed delays as explicit sleep commands with a comment as to why

FreddieAkeroyd (approved):

PR #419, opened 2021-02-18
Use inbuilt asserts to speed up tests

DominicOram:
To test:
* Run the keithley DEVSIM tests on master, confirm they take ~900s
* Run these and confirm they take ~100s and still pass


rerpha (approved):
Much quicker, cheers

PR #426, opened 2021-02-25
adding linmot tests back in

rerpha:
Using context manager for log file as it wasn't being flushed, adding more logging to see if the linmot falls over


DominicOram (approved):
Fingers crossed!

PR #435, opened 2021-03-09
Fix Keithley 2700: Speed up tests and clear buffer as final thing in set up

DominicOram:
To test: 
* Run the tests and confirm they pass (with https://github.com/ISISComputingGroup/EPICS-DeviceEmulator/pull/107)


Tom-Willemsen (approved):
test passes

PR #438, opened 2021-03-16
rename linmot.py to Aalinmot.py to make it run first - need to make s…

rerpha:
…ure we change this back so will write on the ticket


FreddieAkeroyd:
I think it needs more "a"s in the filename :-D

PR #446, opened 2021-04-08
Ticket 6333  - Check for untested iocs

aaron-long:
Added a '-rc', '--report-coverage' command line argument  to `run_tests.py` that will result in the script outputting a list of ioc directories that have not been loaded during the scripts run.

There are a number of ioc test modules which have a 'directory' property. These are used in the  `make_device_launchers_from_module` to create a set of directories loaded during the scripts execution (one or many tests). This set is then compared (in the 'report_test_coverage_for_devices' method) against all folders in the 'EPICS/ioc/master' directory and output.

e.g. run `python run_tests.py -t wbvalve aldn1000 keylkg -rc` and the final list will contain all IOC directories excluding those 3.


run_tests.py
```diff
@@ -156,11 +159,16 @@ def load_and_run_tests(test_names, failfast, ask_before_running_tests, tests_mod
 
         for module in modules_to_be_tested_in_current_mode:
             clean_environment()
-            device_launchers = make_device_launchers_from_module(module.file, mode)
+            device_launchers, device_directories = make_device_launchers_from_module(module.file, mode)
+            if report_coverage:
```
DominicOram, line 163: Let's just always update it, not like it costs anything

run_tests.py
```diff
@@ -186,6 +194,39 @@ def prompt_user_to_run_tests(test_names):
...
+    Report the ioc directories not tested
+
+    Args:
+        tested_directories (list): List of IOC boot directories generated by make_device_launchers_from_module
+
+    Returns:
+        None
+    """
+    # get names of iocs from ioc folder
+    iocs_dir = os.path.join(EPICS_TOP, "ioc", "master")
```
DominicOram, line 208: This is used in a few places, can you make it a constant in ioc_launcher

run_tests.py
```diff
@@ -186,6 +194,39 @@ def prompt_user_to_run_tests(test_names):
...
+    iocs = []
+    for dir in os.listdir(iocs_dir):
+        if os.path.isdir(os.path.join(iocs_dir, dir)):
+            iocs.append(dir)
+    iocs = [ioc.lower() for ioc in iocs]
```
DominicOram, line 213: Could just do this here so you don't need to convert to a set later
```suggestion
    iocs = set(ioc.lower() for ioc in iocs)
```

run_tests.py
```diff
@@ -186,6 +194,39 @@ def prompt_user_to_run_tests(test_names):
...
+        tested_iocs.append(os.path.normpath(dir).split(os.path.sep)[-3].lower())
+
+    iocs = set(iocs)
+    tested_iocs = set(tested_iocs)
+    missing_tests = sorted(iocs.difference(tested_iocs))
+
+    print("\nThe following IOCs have not been tested:\n")
+    for test in missing_tests:
+        print(test)
+    return
```
DominicOram, line 227: No need for a return statement if you're not returning anything

run_tests.py
```diff
@@ -268,6 +309,8 @@ def run_tests(prefix, module_name, tests_to_run, device_launchers, failfast_swit
         description='Test an IOC under emulation by running tests against it')
     parser.add_argument('-l', '--list-devices',
                         help="List available devices for testing.", action="store_true")
+    parser.add_argument('-rc', '--report-coverage',
+                        help='Report devices that have no associated tests.', action="store_true")
```
DominicOram, line 313:
```suggestion
                        help='Report devices that have not been tested.', action="store_true")
```

DominicOram (dismissed):
Generally works well, thanks. Mostly nit-picking things in the code but can you also:
* Update `run_all_tests.bat` so that it runs this
* Update the readme
* Create a PR on the release notes https://github.com/ISISComputingGroup/IBEX/pulls

JamesKingWork (approved):
Looks good, after testing with a few IOCs it produces correct results

PR #449, opened 2021-04-15
ttiex355p: Test bug fixes

JamesKingWork:
ttiex355p: 

- Correct to add minimum voltage and current 
- Test setting setpoint

https://github.com/ISISComputingGroup/IBEX/issues/6397


tests/ttiex355p.py
```diff
@@ -10,8 +10,14 @@
...
-AMPSTOGAUSS = 5
-MAX_FIELD = 10
+AMPSTOGAUSS = 7.278
+MAX_FIELD = 25
+
+# Taken from manual Specification of OUTPUT page 3^M
+MIN_VOLTAGE = 0.0
+MIN_CURRENT = 0.01
+MAX_VOLTAGE = 35.0
+MAX_CURRENT = 5.0
```
DominicOram, line 20: If these are inherent to the device then I think they should just be set in the emulator as default rather than via the backdoor.

tests/ttiex355p.py
```diff
@@ -65,25 +73,25 @@ def test_WHEN_field_set_THEN_current_setpoint_updates(self, _, field):
         self._turn_on_in_const_current_mode()
 
         self.ca.set_pv_value("FIELD:SP", field)
-        self.ca.assert_that_pv_is_number("CURRENT:SP", field / AMPSTOGAUSS, tolerance=0.01)
+        self.ca.assert_that_pv_is_number("CURRENT:SP", field / AMPSTOGAUSS, tolerance=0.08)
```
DominicOram, line 76: Can you move the tolerance out to a constant as it seems to be used in multiple places? Also, why have we modified it?

DominicOram (changes requested):
Tests work and cover the functionality, minor comments

tests/ttiex355p.py
JamesKingWork: They are inherent to the ttiex355p but not the tti355, as such I have moved them into a separate emulator that mostly sub-classes classes of the tti355 with the addition of min voltage and current checks

tests/ttiex355p.py
JamesKingWork: Now moved out to constant. The tolerance has changed because the AMPSTOGAUSS value has changed and division in the IOC is less accurate at this number of decimal places. I have changed the AMPSTOGAUSS value because it more closely represents what is used on instruments

tests/ttiex355p.py
JamesKingWork: Have kept the constants here for use in the tests elsewhere

DominicOram (approved):
Tests pass and look good.

PR #458, opened 2021-05-20
fix the tests broken by the reverted lewis backdoor control

rerpha:
I have run all of these tests, you could do but it'll take you a while, probs best to let the build server do it!


DominicOram (approved):
Looks good

PR #459, opened 2021-05-20
Ticket6256 hanging ioc tests

rerpha:


DominicOram (approved):
Changes look good, thanks

PR #460, opened 2021-05-24
fix more tests after lewis backdoor control change

rerpha:
fixes danfysik8800,8500 and common tests and rotsc tests which were broken after the lewis backdoor reversion


DominicOram (approved):
Great, thanks

PR #463, opened 2021-05-28
add sleep for zfcntrl autosave test - not getting saved quickly enough

rerpha:


DominicOram (approved):
Annoying we have to put a sleep in but looks like it will make it more stable

PR #468, opened 2021-06-15
fix zfcntrl autofeedback autosave test and speedup

rerpha:


DominicOram (approved):
Looks good and I saw it fixing the tests on the build server

PR #472, opened 2021-06-17
adding test for creep speed - second test still broken

rerpha:


rerpha:
second test wasn't broken, my build was, they should all pass

PR #473, opened 2021-06-23
add retry when stopping motors in refl tests

ThomasLohnert:
`MOT:STOP:ALL` sometimes fails to stop all axes when on simulated motors. This PR adds a retry to try to make this less likely to break our tests.

### DIFF
--- a/tests/refl.py
+++ b/tests/refl.py
@@ -132,6 +132,22 @@ def all_motors_in_set_mode(channel_access):
             channel_access.set_pv_value(offset_freeze_switch_pv.format(motor), offset_freeze_switch)
 
 
+def stop_motors_with_retry(channel_access, n_retry):
+    """
+    Tries to stop all motors several times as this operation sometimes fails for simulated axes.
+
+    Args:
+     channel_access: The channel access interface to use
+     n_retry: The number of retries
+    """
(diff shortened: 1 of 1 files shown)


tests/refl.py
```diff
@@ -132,6 +132,22 @@ def all_motors_in_set_mode(channel_access):
...
+    """
+    Tries to stop all motors several times as this operation sometimes fails for simulated axes.
+
+    Args:
+     channel_access: The channel access interface to use
+     n_retry: The number of retries
+    """
+    for _ in range(n_retry):
+        channel_access.set_pv_value("MOT:STOP:ALL", 1, wait=True)
+        if channel_access.get_pv_value("MOT:MOVING") != 0.0:
```
RaiBishal, line 145: If I am not mistaken, we are checking if the motors are still moving here? Should this not be `channel_access.get_pv_value("MOT:MOVING") != 1.0:` instead? I am assuming `get_pv_value("MOT:MOVING")` will be`1.0` if motors are moving and `0.0` if the motors are stopped.

ThomasLohnert: whoops, you're absolutely right

RaiBishal (approved):

PR #476, opened 2021-07-26
Added multiple tests for ReadASCII, primarily focusing on file and reading and crash prevention

Adam-Szw:
[Ticket 6352](https://github.com/ISISComputingGroup/IBEX/issues/6352)


Adam-Szw:
Code PR: https://github.com/ISISComputingGroup/EPICS-ReadASCII/pull/14

FreddieAkeroyd:
Thanks for adding more tests - it would also be useful to test behaviour when:
- a column header name is repeated
- currently the data file always has the same name and is reloaded, could changing filename be tested too
- could setting a non-existent filename and then setting a valid one be checked for behaviour e.g. does it initially alarm the filename PV and then load things ok later

PR #480, opened 2021-08-10
Fix mercury tests

FreddieAkeroyd:
Fix mercury tests

we have a lot of db records scanning in the ioc, they will start to fail but it may take a while (and this will be variable) before the record we choose gets processed and fails or hits the stream device lock timeout instead. So we need to wait for at least as long as stream device lock timeout to see an alarm raised

### DIFF
--- a/tests/mercuryitc.py
+++ b/tests/mercuryitc.py
@@ -508,7 +508,11 @@ class MercuryTests(unittest.TestCase):
         expected_value = -10
         self.ca.set_pv_value("{}:PRESSURE:SP".format(pressure_card_pv_prefix), expected_value)
         self._lewis.backdoor_set_on_device("connected", False)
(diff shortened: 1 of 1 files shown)


JamesKingWork:
Should we not be fixing the underlying issue and making the db records scanning better?

FreddieAkeroyd:
fair point - I took that to be a separate issue that would require more testing, probably with hardware. The problem here was that the test was not correct for the way the ioc currently works, so I fixed the test so it was. It probably worked ok when lock timeout was the default 5 seconds. even with chaining records rather than scanning individually for example you may get a similar issue, if the connection is dropped mid scan chain it suddenly takes a long time to process rest of chain, and if you have looked at start of chain for an alarm you may miss it if your timeout is too small.

PR #485, opened 2021-09-16
Disable the unstable danfysik test

rerpha:
Ticket to fix and re enable it - https://github.com/ISISComputingGroup/IBEX/issues/6787


FreddieAkeroyd:
Could we use a decorator rather than commenting out, like we do with `skip_if_recsim` ?

rerpha:
Closing as the test is now fixed

PR #487, opened 2021-09-20
Fix log file reading - think previous approach could miss messages

FreddieAkeroyd:
I think the previous approach did some unnecessary seeking and may sometime miss a message. This is an attempt to fix.


JamesKingWork:
This seems to cause the zfcntrl tests to fail to start the kepcos for me

FreddieAkeroyd:
if you tell it to match on `iocRun: All initialization complete` rather than `epics>` does that change anything?

PR #508, opened 2021-11-02
removed old tests

aaron-long:


rerpha:
Just going to close this as we should remove the astrium IOC instead (linked in ticket)

PR #526, opened 2022-04-06
Ticket5877 Added more tests to tpg300 to test new functionality. Modified some old tests

Adam-Szw:
https://github.com/ISISComputingGroup/IBEX/issues/5877

### DIFF
--- a/tests/tpg300.py
+++ b/tests/tpg300.py
@@ -1,11 +1,14 @@
 import unittest
 
+from parameterized import parameterized
+
 from utils.channel_access import ChannelAccess
 from utils.ioc_launcher import get_default_ioc_dir
 from utils.test_modes import TestModes
 from utils.testing import get_running_lewis_and_ioc, skip_if_recsim
 from enum import Enum
 from itertools import product
+from time import sleep
 
 DEVICE_PREFIX = "TPG300_01"
 
@@ -52,7 +55,7 @@ class Tpg300Tests(unittest.TestCase):
 
     def _set_units(self, unit):
         self._lewis.backdoor_run_function_on_device("backdoor_set_unit", [unit.value])
-        self._ioc.set_simulated_value("SIM:UNITS", unit.name)
+        self.ca.set_pv_value("UNITS:SP:RBV", unit.name)
 
     def _connect_emulator(self):
         self._lewis.backdoor_run_function_on_device("connect")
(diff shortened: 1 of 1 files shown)


tests/tpg300.py
```diff
@@ -52,7 +55,7 @@ def _set_pressure(self, expected_pressure, channel):
 
     def _set_units(self, unit):
         self._lewis.backdoor_run_function_on_device("backdoor_set_unit", [unit.value])
-        self._ioc.set_simulated_value("SIM:UNITS", unit.name)
+        self.ca.set_pv_value("UNITS:SP:RBV", unit.name)
```
Tom-Willemsen, line 58: Writing to a setpoint readback doesn't usually make sense... a setpoint readback should be read directly from the hardware/emulator.

Can we set the `SIM` pv instead if this is only for recsim?

tests/tpg300.py
```diff
@@ -83,3 +87,70 @@ def test_that_GIVEN_a_disconnected_emulator_WHEN_getting_pressure_THEN_INVALID_a
...
+        self.ca.set_pv_value("FUNCTION:HIGH:E:SP", exponent_high)
+        self.ca.set_pv_value("FUNCTION:ASSIGN:SP", circuit_assignment)
+        self.ca.process_pv("FUNCTION:ASSIGN:SP:OUT")
+
+    def check_switching_function_thresholds(self, function, thresholds, check_pv_is=True):
+        readback_string = str(thresholds[0]) + 'E' + str(thresholds[1]) + ',' + str(thresholds[2]) + 'E' + \
+                          str(thresholds[3]) + ',' + str(thresholds[4])
+        if check_pv_is:
+            self.ca.assert_that_pv_is("FUNCTION:"+function+":RB", readback_string)
+            self.ca.assert_that_pv_is("FUNCTION:"+function+":LOW:RB", thresholds[0])
```
Tom-Willemsen, line 107: For floating-point values use `assert_that_pv_is_number` with an appropriate tolerance to avoid rounding errors causing test failures.

tests/tpg300.py
```diff
@@ -83,3 +87,70 @@ def test_that_GIVEN_a_disconnected_emulator_WHEN_getting_pressure_THEN_INVALID_a
...
+            self.ca.assert_that_pv_is("FUNCTION:"+function+":LOW:E:RB", thresholds[1])
+            self.ca.assert_that_pv_is("FUNCTION:"+function+":HIGH:RB", thresholds[2])
+            self.ca.assert_that_pv_is("FUNCTION:"+function+":HIGH:E:RB", thresholds[3])
+            self.ca.assert_that_pv_is("FUNCTION:"+function+":ASSIGN:RB", thresholds[4])
+        else:
+            self.ca.assert_that_pv_is_not("FUNCTION:"+function+":RB", readback_string)
+            self.ca.assert_that_pv_is_not("FUNCTION:"+function+":LOW:RB", thresholds[0])
```
Tom-Willemsen, line 114: Similar to above, there is an `assert_that_pv_is_not_number` with a tolerance which is more suitable for floating-point values.

PR #553, opened 2022-08-08
Add Percent of Volt Limit Tests

ConorMFinn:
Add some tests for the temp and pressure cards to make sure that the calculation is happening as expected. 

Partially Resolves ISISComputingGroup/IBEX#6283


JackEAllen (approved):
Looks good and happy to merge in. But wanted to ask how you feel about the reuse of setting:
```python
original_voltage = 10
new_voltage = 20
voltage_limit = 50
original_percent = 20
new_percent = 40
```
in both tests? Do you think it's worth us moving this into `__init__` and calling from there so we only initialise once? 

I don't mind either way, but wanted to ask in case you agree.

PR #557, opened 2022-08-17
Ticket7276 constrain log size

NikolaRoev:
https://github.com/ISISComputingGroup/IBEX/issues/7276

### DIFF
--- a/test_config/good_for_refl/refl/config_trunc.py
+++ b/test_config/good_for_refl/refl/config_trunc.py
@@ -7,7 +7,7 @@ import random
 import string
 from ReflectometryServer import *
 
-
+random.seed(1)
 RANDOM_STRING = ''.join(random.choices(string.ascii_letters + string.digits, k=5000))
 
 
--- a/test_config/good_for_refl/refl/config_trunc.py
+++ b/test_config/good_for_refl/refl/config_trunc.py
@@ -1,0 +1,33 @@
+"""
+FOR TESTING
+
+Valid configuration script for a reflectometry beamline
+"""
+import random
+import string
+from ReflectometryServer import *
+
+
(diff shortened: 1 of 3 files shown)


test_config/good_for_refl/refl/config_trunc.py
```diff
@@ -0,0 +1,33 @@
...
+FOR TESTING
+
+Valid configuration script for a reflectometry beamline
+"""
+import random
+import string
+from ReflectometryServer import *
+
+
+RANDOM_STRING = ''.join(random.choices(string.ascii_letters + string.digits, k=5000))
```
Tom-Willemsen, line 11: Please seed the random number generator if you absolutely must use `random`, to ensure we don't get random failures that are very hard to debug

tests/refl_trunc.py
```diff
@@ -0,0 +1,122 @@
...
+        self.assertEqual(size, max_size, f"PARAM_INFO size {size} should be {max_size}.")
+
+    def test_WHEN_server_message_exceeds_character_limit_THEN_message_truncated_correctly(self):
+        SERVER_MESSAGE_MAX_SIZE = 400
+
+        self.ca.set_pv_value("PARAM:INBEAM:SP", 0, wait=True)
+        sleep(2)
+
+        message : str = self.ca.get_pv_value("MSG")
+        self.assertTrue(message.endswith("<truncated>"), f"Server message has to end with '<truncated>'.\nMESSAGE:\n{message}")
```
Tom-Willemsen, line 103: Instead of sleeping, you can use `assert_that_pv_value_causes_func_to_return_true`, and pass in `lambda val: val.endswith('truncated')` as the function.

This has the benefit that the assertion will pass as soon as the condition is true, and won't always have to sleep for 2s. Generally we should pretty much never need `sleep` in our IOC tests.

Tom-Willemsen (approved):

PR #558, opened 2022-08-17
Refactor IOC test framework device disconnection tests & add README documentation

boeuf-codes:
**Ticket: https://github.com/ISISComputingGroup/IBEX/issues/7266**
Refactoring all test modules which contain the common behaviour as described in the ticket.

- **Added documentation in IOC Test Framework README for use of new util fn**
- **Added a emulator util function to generalise common device disconnection behaviour**
- Used new emulator util function instead of old logic
- Added in asserts/before main assertions for robustness
- Occasionally fixed some tests which threw errors as a result of the above


JackEAllen (changes requested):
This was a big one! Thanks for tackling it! 👍 
Just left minor comments here and there, feel free to challenge me on any or them 🙂
I'll place this into rework for the time being until I get your opinion on my comments.

JackEAllen (approved):
LGTM! 👍

PR #566, opened 2022-09-26
Ticket7376

NikolaRoev:
https://github.com/ISISComputingGroup/IBEX/issues/7376

### DIFF
--- a/tests/keylkg.py
+++ b/tests/keylkg.py
@@ -5,7 +5,7 @@ from parameterized import parameterized
 from utils.channel_access import ChannelAccess
 from utils.ioc_launcher import get_default_ioc_dir
 from utils.test_modes import TestModes
-from utils.testing import get_running_lewis_and_ioc, skip_if_recsim
+from utils.testing import get_running_lewis_and_ioc, skip_if_recsim, parameterized_list
 
 
 DEVICE_PREFIX = "KEYLKG_01"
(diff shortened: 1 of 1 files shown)


tests/keylkg.py
```diff
@@ -146,3 +146,12 @@ def test_GIVEN_running_ioc_WHEN_in_setup_mode_THEN_output1_switches_to_measureme
         self._lewis.backdoor_set_on_device("detector_1_raw_value", expected_value)
 
         self.ca.assert_that_pv_is("VALUE:OUTPUT:1", expected_value)
+
+    @parameterized.expand(parameterized_list(["Passive", "5 second"]))
+    @skip_if_recsim("Scan rate changes do not work in RECSIM.")
+    def test_WHEN_scan_rates_changed_THEN_scan_rates_set_correctly(self, _, rate):
```
Tom-Willemsen, line 152: There is a test-ordering condition in these tests; you can test this by replacing

```
unittest.TestLoader().loadTestsFromNames(test_names)
```

with 

```
    loader = unittest.TestLoader()
    loader.sortTestMethodsUsing = lambda *a, **k: 1
    test_suite = loader.loadTestsFromNames(test_names)
```

in `run_tests.py`, this makes the tests run in a different order and then a load of them fail. The best fix for this is probably to set the `SCAN` back to `5 second` in `setUp()`.

Tom-Willemsen (changes requested):

Tom-Willemsen (approved):

PR #590, opened 2023-02-24
Add tests for extra moving indicator record

boeuf-codes:


tests/oscillating_collimator_let.py
```diff
@@ -24,7 +25,7 @@
         },
     },
 ]
-TEST_MODES = [TestModes.DEVSIM]
+TEST_MODES = [TestModes.RECSIM]
```
rerpha, line 28: we might be able to run these tests in devsim and recsim, but just run the ones that require digital input signals in recsim

common_tests/oscillating_collimators.py
```diff
@@ -6,7 +6,8 @@
 DISTANCE = "DIST:SP"
 DISCRIMINANT = "VEL:SP:DISC:CHECK"
 GALIL_ADDR = "127.0.0.11"
-PREFIX = "MOT:OSCCOL"
+MOT_PREFIX = "MOT"
+OSC_PREFIX = MOT_PREFIX + ":OSCCOL"
```
Tom-Willemsen, line 10: wish collimator tests are now broken because this variable name has changed.

tests/oscillating_collimator_let.py
Tom-Willemsen: As far as I know, "recsim" and "devsim" are actually the same thing for a galil?

tests/oscillating_collimator_let.py
```diff
@@ -83,3 +85,22 @@ def test_WHEN_input_values_cause_discriminant_to_be_negative_THEN_discriminant_p
 
         # Assert
         self.ca.assert_that_pv_is_number(DISCRIMINANT, 1.0)
+
+    def test_WHEN_laser_unstable_THEN_collimator_is_reported_moving(self):
+        # Act
+        # in normal operations the radius is not dynamic so set it first so it is considered in future calcs
+        # Set 0 first then 1 so buffer contents are definitively inconsistent
+        self.ca_mot.set_pv_value("DMC01:Galil0Bi5_STATUS", 0)
+        self.ca_mot.set_pv_value("DMC01:Galil0Bi5_STATUS", 1)
```
Tom-Willemsen, line 94: Since your buffer scans once a second, if it was already full of 1s you may miss the intermediate `0` state?

Best to wait a short time (e.g. 1s) between these two sets, or actually monitor the buffer and make sure that it has different values in it (you may need to use `assert_pv_value_causes_func_to_return_true`, or a similar custom assertion, for this).

tests/oscillating_collimator_let.py
```diff
@@ -83,3 +85,22 @@ def test_WHEN_input_values_cause_discriminant_to_be_negative_THEN_discriminant_p
...
+
+        # Assert
+        self.ca.assert_that_pv_is("COLLIM_MOVING", "MOVING")
+    
+    def test_WHEN_laser_stable_THEN_collimator_is_reported_stopped(self):
+        # Act
+        # in normal operations the radius is not dynamic so set it first so it is considered in future calcs
```
Tom-Willemsen, line 101: ? Doesn't look relevant to this test?

tests/oscillating_collimator_let.py
```diff
@@ -83,3 +85,22 @@ def test_WHEN_input_values_cause_discriminant_to_be_negative_THEN_discriminant_p
 
         # Assert
         self.ca.assert_that_pv_is_number(DISCRIMINANT, 1.0)
+
+    def test_WHEN_laser_unstable_THEN_collimator_is_reported_moving(self):
+        # Act
+        # in normal operations the radius is not dynamic so set it first so it is considered in future calcs
```
Tom-Willemsen, line 91: ? Doesn't look relevant to this test?

tests/oscillating_collimator_let.py
```diff
@@ -83,3 +85,22 @@ def test_WHEN_input_values_cause_discriminant_to_be_negative_THEN_discriminant_p
...
+        self.ca.assert_that_pv_is("COLLIM_MOVING", "MOVING")
+    
+    def test_WHEN_laser_stable_THEN_collimator_is_reported_stopped(self):
+        # Act
+        # in normal operations the radius is not dynamic so set it first so it is considered in future calcs
+        # Will take ten seconds for buffer to have entirely consistent contents
+        self.ca_mot.set_pv_value("DMC01:Galil0Bi5_STATUS", 1, sleep_after_set=10)
+
+        # Assert
+        self.ca.assert_that_pv_is("COLLIM_MOVING", "STOPPED")
```
Tom-Willemsen, line 106: Perhaps better to assert that the pv _was_ MOVING, and then becomes (and stays) STOPPED after about 10s? I can help you write this if needed.

Tom-Willemsen (approved):

PR #604, opened 2023-07-06
Refactor and generalise TPG300 IOC tests

boeuf-codes:
### Description of work

- Rename module to `tpgx00`, move to common_tests
- Create new submodules `tpg300` and `tpg500`
- Various tidying and refactoring of older tests
- Add tests for added pressure status PVs
- Refactor tests inline with refactored emulator behaviour
- Add tests to check IOC behaviour with invalid commands 
   - Add `assert_that_emulator_value_is_not` method to `utils/emulator_launcher.py` to support this
- Add test for ERR command
- Move some general tests to `tpg300` whilst hardware issues with the TPG500 are being sorted (https://github.com/ISISComputingGroup/IBEX/issues/7860)

### Ticket

[#7458](https://github.com/ISISComputingGroup/IBEX/issues/7458)

### DIFF
--- a/common_tests/tpgx00.py
+++ b/common_tests/tpgx00.py
@@ -16,7 +16,7 @@ class ChannelStatus(Enum):
     DATA_OK     = ("Measured data okay", "NO_ALARM")
     UNDERRANGE  = ("Underrange", "MINOR")
     OVERRANGE   = ("Overrange", "MINOR")
-    POINT_ERROR   = ("Point error", "MAJOR")
+    POINT_ERROR = ("Point error", "MAJOR")
     POINT_OFF   = ("Point switched off", "MAJOR")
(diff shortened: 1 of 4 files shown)


zsoltkebel:
I removed the test `tpg300.test_WHEN_invalid_unit_set_THEN_pv_goes_into_alarm` because couldn't get the `mbbo` record on `UNITS:SP` to go into INVALID alarm when set out of range. Not sure if this kind of testing is common for mbbo-s in other IOCs. 

This was a newly added test which seems to have been pointless as the PV was just in INVALID state because it did not have a value initialised.

Tried setting pv to an accepted value in the first line made it fail too. (Should still work as afterwards it is set to an invalid unit which should raise an alarm)

esmith1729 (approved):
tests passed

PR #607, opened 2023-08-18
Adding tests for the 1242

KathrynBaker:


esmith1729 (approved):
DI test will need to be fixed in https://github.com/ISISComputingGroup/IBEX/issues/7963

PR #608, opened 2023-09-04
Set REMOTE_ON_SET to NO

FreddieAkeroyd:
See ISISComputingGroup/IBEX#7778


tests/kepco_no_rem.py
```diff
@@ -16,7 +16,7 @@
     {
         "name": DEVICE_PREFIX,
         "directory": get_default_ioc_dir("KEPCO"),
-        "macros": {"CURRENT_MAX": MAX_CURRENT},
+        "macros": {"CURRENT_MAX": MAX_CURRENT, "REMOTE_IN_SET": "NO"},
```
rerpha, line 19: i think this should be REMOTE_ON_SET - should kepco_rem also have this?

rerpha, line 19:
```suggestion
        "macros": {"CURRENT_MAX": MAX_CURRENT, "REMOTE_ON_SET": "NO"},
```

FreddieAkeroyd: Thanks for spotting typo, REMOTE_ON_SET defaults to YES so kepco_rem doesn't need it to run correctly

rerpha (approved):

PR #613, opened 2023-10-10
Record host name separately in channel access

zsoltkebel:


esmith1729 (approved):
Appears to not affect existing functionality, while allowing for the edge case needed for the RIKEN monitor.

PR #620, opened 2024-04-15
Ensure subprocess termination on python exit

FreddieAkeroyd:
See ISISComputingGroup/IBEX#7957


esmith1729 (approved):
Works as expected after following instructions from https://github.com/ISISComputingGroup/IBEX/issues/7957

PR #621, opened 2024-04-20
Allow for exiting tests before tests to handle cleanup of ioc

FreddieAkeroyd:


utils/device_launcher.py
```diff
@@ -26,7 +26,11 @@ def device_collection_launcher(devices):
...
-        for device in devices:
-            stack.enter_context(device)
-        yield
+    try:
+        with ExitStack() as stack:
+            for device in devices:
+                stack.enter_context(device)
+            yield
+    except GeneratorExit:
+        print("Cleaning up... please wait")
```
FreddieAkeroyd, line 35: I never see this message printed from the call to `device_launchers.__exit__(None, None, None)` - do you? I tried adding a `file=sys.stderr` (though as we are running unbuffered output that should be happening anyway).  The cleanups run anyway after the call to `sys.exit(0)` so i'm not sure is we need to call  `device_launchers.__exit__`.  I tried adding an `except SystemExit` to print a message like above and it did print one from using just `sys.exit(0)` but that message appeared right at the end after all other messages and so not so useful.

isaachilly: Okay, yes, I put the except in as just another way to catch the program exiting and alerting user but I think I misunderstood the mechanics. I just tested without __exit__ and yes also works. I must've added in at start and then assumed that's been calling the closures rather than the SystemExit exception.

isaachilly:
Made a couple of changes.

FreddieAkeroyd (commented):
Thank - looks great

PR #625, opened 2024-05-13
Refactoring eurotherm tests to work with multiple sensors

esmith1729:
Added unit test for new functionality created from 

- https://github.com/ISISComputingGroup/EPICS-ioc/pull/851

can be tested with: 

```python -u run_tests.py -t eurotherm_eibisynch.EurothermTests.test_WHEN_eurotherm_missing_THEN_updates_of_PVs_stop   ```


esmith1729:
Linter is failing for `utils\emulator_launcer.py`, this should be reformatted during a "Ruff day", it will also need some changes due to pyright, I would guess, but that falls outside of scope of this ticket.

Chsudeepta (approved):

PR #639, opened 2024-09-17
add test for muon tpar editor

rerpha:


Chsudeepta (commented):
There are some errors reported by Ruff. It seems there are some unused imports.

Chsudeepta (commented):
1. It seems that the save test case should fail in general since the manager mode should be On for the save action.
2. Ideally there should be two assertions, one setting the manager mode off and assert failure to save and the other setting the manager mode On and asserting success to save
3. It seems that the Readd test case is failing the assertion due to extra CRLF.

Chsudeepta (approved):
working as expected

PR #643, opened 2024-10-04
Ticket8359 create p4p wrapper

LowriJenkins:
https://github.com/ISISComputingGroup/IBEX/issues/8359

Allowed the use of P4P in the ioc test framework and set the galil tests to use them for testing purposes.

## to test
- [ ] run the galil tests.
- [ ] in another test set the `pv_access` parameter on the `ChannelAccess` object to true to attempt using P4P instead of channel access in that test.


rerpha:
just a general comment: 
I would quite like to make PVA use easily configurable from [run_tests.py](https://github.com/ISISComputingGroup/EPICS-IOC_Test_Framework/blob/master/run_tests.py) or from an environment variable so we can run _all_ of the tests against PVA. could the `if` statements honour something like this for choosing whether to use CA or PVA? 
we should make a ticket once this is done to set up a jenkins job to do it

Tom-Willemsen:
I think Jack has a valid point here - though think I'd slightly favour a command line flag to the test framework for default CA library, but substantially the same thing

Pushed small changes to fix imports.

utils/channel_access.py
```diff
@@ -7,36 +7,36 @@
...
+from functools import partial, partialmethod
+from typing import Callable, Generator, Optional
 
-from genie_python.genie_cachannel_wrapper import CaChannelWrapper, UnableToConnectToPVException
+from genie_python.genie import PVValue  # type: ignore
+from genie_python.genie_cachannel_wrapper import (  # type: ignore
+    CaChannelWrapper,
+    UnableToConnectToPVException,
+)
+from genie_python.genie_p4p_wrapper import P4PWrapper  # type: ignore
```
Tom-Willemsen, line 20: btw, you should now be able to add `genie_python>=15.1.0-rc1` to `requirements.txt` and pyright on the build server should then pick everything up correctly 🤞 - i.e. you should hopefully not need the `type: ignore` any more...

Tom-Willemsen:
Have taken the liberty to push https://github.com/ISISComputingGroup/EPICS-IOC_Test_Framework/pull/643/commits/ec0bc1e297325f4452ac16ecfa4322a169bb5e30 onto your branch, as otherwise pva was never used

Tom-Willemsen (approved):

PR #654, opened 2025-02-10
Add longer timeout to alarm none check, add robust checking to eurotherm missing test

esmith1729:
Fix to System Tests intermittently failing when alarm was expected to be NONE returned as INVALID due to not enough time given for Alarm PV to return.


esmith1729:
Please put SPARE70 node back online after reviewing this as well, thank you

PR #672, opened 2026-06-01
Launch cygserver in tests to avoid long pauses with procServ

FreddieAkeroyd:
Launch cygserver to avoid slow test running on developer computers. System test servers are currently unaffected by this issue due to the account they run under.


Chsudeepta (approved):
Tested. Working fine

PR #673, opened 2026-06-05
Added test cases for new Eurotherm records

Chsudeepta:
This is related with the [ticket](https://github.com/ISISComputingGroup/IBEX/issues/8688)

### DIFF
--- a/common_tests/eurotherm.py
+++ b/common_tests/eurotherm.py
@@ -430,7 +430,6 @@ class EurothermBaseTests(
         self.ca.assert_that_pv_is_number("A01:LOWLIM", val, tolerance=0.05, timeout=15)
 
     @parameterized.expand(parameterized_list(BINARY_VALUES))
-    @skip_if_recsim("Backdoor not available in recsim")
     def test_WHEN_automan_set_via_backdoor_THEN_automan_updates(self, _, val):
         expected_value = {0: "OFF", 1: "ON"}[val]
 
--- a/common_tests/eurotherm.py
+++ b/common_tests/eurotherm.py
@@ -441,7 +441,6 @@ class EurothermBaseTests(
     @skip_if_recsim("Backdoor not available in recsim")
     def test_WHEN_sensorbreak_set_via_backdoor_THEN_shows_updated_value(self, _, val):
         expected_value = {0: "OFF", 1: "ON"}[val]
-        
+
         self._lewis.backdoor_run_function_on_device("set_snbrkpst", [SENSORS[0], val])
         self.ca.assert_that_pv_is("A01:SNBRKPST", expected_value)
-
--- a/common_tests/eurotherm.py
+++ b/common_tests/eurotherm.py
@@ -27,6 +27,8 @@ SENSORS = ["01", "02", "03", "04", "05", "06"]
 
 PV_SENSORS = ["A01", "A02", "A03", "A04", "A05", "A06"]
 
+BINARY_VALUES = [0, 1]
+
 
 # This class is only valid for classes which also derive from unittest.TestCase,
 # and we can't derive from unittest.TestCase at runtime, because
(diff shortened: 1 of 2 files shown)


common_tests/eurotherm.py
```diff
@@ -418,3 +426,19 @@ def test_WHEN_high_limit_set_via_backdoor_THEN_high_lim_updates(self, _, val):
     def test_WHEN_low_limit_set_via_backdoor_THEN_low_lim_updates(self, _, val):
         self._lewis.backdoor_run_function_on_device("set_low_lim", [SENSORS[0], val])
         self.ca.assert_that_pv_is_number("A01:LOWLIM", val, tolerance=0.05, timeout=15)
+
+    @parameterized.expand(parameterized_list(TEST_VALUES))
+    @skip_if_recsim("Backdoor not available in recsim")
```
FreddieAkeroyd: This test isn't using teh lewis backdoor and so shoudl be able to be run in recsim

common_tests/eurotherm.py
```diff
@@ -418,3 +426,19 @@ def test_WHEN_high_limit_set_via_backdoor_THEN_high_lim_updates(self, _, val):
...
+    def test_WHEN_automan_set_via_backdoor_THEN_automan_updates(self, _, val):
+        self.ca.set_pv_value("A01:AUTOMAN", 1)
```
FreddieAkeroyd: The test should be setting `AUTOMAN:SP` rather than `AUTOMAN`. Also you coudl make the parameterised list have `[0,1]` and then use `val` in the test

common_tests/eurotherm.py
```diff
@@ -418,3 +426,19 @@ def test_WHEN_high_limit_set_via_backdoor_THEN_high_lim_updates(self, _, val):
...
+        self.ca.assert_that_pv_is("A01:AUTOMAN", "ON")
+        self.ca.set_pv_value("A01:AUTOMAN", 0)
+        self.ca.assert_that_pv_is("A01:AUTOMAN", "OFF")
+
+    @parameterized.expand(parameterized_list(TEST_VALUES))
+    @skip_if_recsim("Backdoor not available in recsim")
+    def test_WHEN_sensorbreak_set_via_backdoor_THEN_shows_updated_value(self, _, val):
```
FreddieAkeroyd, line 441: The test isn't currently using the backdoor - it should use the lewis backdoor to change the emulator snbrkpst value to `val` and then check the readback in A01:SNBRKPST - similar to what you have alrady done for the low and high limit tests

common_tests/eurotherm.py
```diff
@@ -418,3 +426,19 @@ def test_WHEN_high_limit_set_via_backdoor_THEN_high_lim_updates(self, _, val):
...
+
+    @parameterized.expand(parameterized_list(TEST_VALUES))
+    @skip_if_recsim("Backdoor not available in recsim")
+    def test_WHEN_automan_set_via_backdoor_THEN_automan_updates(self, _, val):
+        self.ca.set_pv_value("A01:AUTOMAN", 1)
+        self.ca.assert_that_pv_is("A01:AUTOMAN", "ON")
+        self.ca.set_pv_value("A01:AUTOMAN", 0)
+        self.ca.assert_that_pv_is("A01:AUTOMAN", "OFF")
+
+    @parameterized.expand(parameterized_list(TEST_VALUES))
```
FreddieAkeroyd: you probably need different values here in the parameterised list e.g. 0 and 1 and ten you can use `val` in your tests - similar to as with `AUTOMAN` above

common_tests/eurotherm.py
Chsudeepta: This is a pre-existing test.

FreddieAkeroyd: what do you mean by "pre-exising test"?

Chsudeepta: Seems I made a wrong comment. I thought you were referring to the test on LOWLIM which isn't part of my changes.

FreddieAkeroyd: Ah ok, no I was referring to the newly added `AUTOMAN` that you should be able to remove the `@skip_if_recsim` from. When i mentioned `LOWLIM` in the `SNBRKPST` saying you should use a backdoor like in this test, I had though you had also added LOWLIM at some point - i know you had added a `LOWRANGE` PV for "Low range limit" and i had though this was a test for that.

FreddieAkeroyd (approved):

PR #674, opened 2026-07-18
Fix settings paths for tests

FreddieAkeroyd:
* Set new SIM value PV in DI tests
* Add missing galil setup files, also change some tests to use main instrument after deleting copies in https://github.com/ISISComputingGroup/EPICS-motorExtensions/pull/87

See https://github.com/ISISComputingGroup/IBEX/issues/8979

### DIFF
--- a/common_tests/DAQmx.py
+++ b/common_tests/DAQmx.py
@@ -8,7 +8,7 @@ DEVICE_PREFIX = "DAQMXTEST"
 ICPCONFIGNAME = "DAQMX"
 
 
-class DAQmxTests(object):
+class DAQmxTests:
     """
     General tests for the DAQmx.
     """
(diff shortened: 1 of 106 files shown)


FreddieAkeroyd:
could of cases i have ignored `BLE001` but maybe could catch  `AssertionError | UnableToConnectToPVException` if that covered things e.g. in cryosms?

FreddieAkeroyd:
just rerunning a jenkins job as seen some  `AttributeError: type object 'datetime.datetime' has no attribute 'timezone'. Did you mean: 'astimezone'?` in an earlier system test log. Feels like you might get this if you had done a `from datetime import datetime` rather than `import datetime` but i'd have thought ruff would have picked this up. Re-running as it was looking at a slightly older version of code, but I don't recall changing any datetime today.

Tom-Willemsen (approved):

